PR #6, opened 2024-05-10
Fix JDBC type for DECIMAL columns; add test

elefeint:
This PR allows `DECIMAL(W,S)` types to be interpreted as `DECIMAL` JDBC type. Currently the type is returned as a generic `JAVA_OBJECT` because  `DECIMAL(W,S)` does not match ` `DECIMAL` in the case statement. This is similar to [0.9.2 behavior](https://github.com/duckdb/duckdb/blob/v0.9.2/tools/jdbc/src/main/java/org/duckdb/DuckDBDatabaseMetaData.java#L832) except the start-of-string matching is done in SQL rather than [Java](https://github.com/duckdb/duckdb/blob/v0.9.2/tools/jdbc/src/main/java/org/duckdb/DuckDBResultSetMetaData.java#L49).

I wonder if a better fix would be to use `data_type_id` from `duckdb_columns()` rather than `data_type` from `information_schema.columns`? It would require adding hardcoded duckdb type numbers into `DuckDBColumnType`. The test is useful either way.

There is a tiny unrelated improvement to `Assertions` utility, allowing the calling code to add a prefix to indicate which one of the loop iterations failed.

Fixes duckdb/duckdb#11365

### DIFF
--- a/src/main/java/org/duckdb/DuckDBDatabaseMetaData.java
+++ b/src/main/java/org/duckdb/DuckDBDatabaseMetaData.java
@@ -825,7 +825,7 @@ public class DuckDBDatabaseMetaData implements DatabaseMetaData {
                                   + "table_catalog AS 'TABLE_CAT', "
                                   + "table_schema AS 'TABLE_SCHEM', "
                                   + "table_name AS 'TABLE_NAME', "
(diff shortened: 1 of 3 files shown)


Mause:
Thanks!

PR #25, opened 2024-06-04
correct error in action name, arm64 -> amd64

rom38:


Mause:
Hi there, did you mean to add two new workflows?

Issue #26, opened 2024-06-06
Metadata Functions for SchemaSpy


joshuashaffer:
I was trying to run SchemaSpy on a DuckDB database with indices defined, and it was dying after throwing a host of exceptions related to missing metadata functions. I took a [stab at implementing the ones it was complaining about](https://github.com/joshuashaffer/duckdb-java-schema-spy-metadata/commit/b416a2ef19df61b5dae9500ecde94c3430fc3ee7), and it seems to work now by producing diagrams. (There are still a few left, but they seem recoverable.)

I'm not a JDBC expert, so I don't know if this is good enough. Is this good enough for a PR, or what else would you like to see?

Mause:
Hey there, those look great - if you're able to write some quick tests I'd be happy to merge those

PR #27, opened 2024-06-08
Schemaspy metadata

joshuashaffer:
Implements a few metadata functions so SchemaSpy mostly works. Issue #26


src/main/java/org/duckdb/DuckDBDatabaseMetaData.java
```diff
@@ -171,27 +171,76 @@ public String getIdentifierQuoteString() throws SQLException {
...
+        return sb.toString();
     }
 
     @Override
     public String getSystemFunctions() throws SQLException {
-        throw new SQLFeatureNotSupportedException("getSystemFunctions");
+        Statement statement = conn.createStatement();
+        statement.closeOnCompletion();
+        ResultSet rs = statement.executeQuery(
+                "SELECT DISTINCT function_name FROM duckdb_functions() WHERE parameter_types[1] IS NULL");
```
Mause, line 223: This condition doesn't quite work - probably need to exclude vararg functions somehow?

Mause (changes requested):

src/main/java/org/duckdb/DuckDBDatabaseMetaData.java
joshuashaffer:
```sql 
select DISTINCT function_name from duckdb_functions() where length(parameter_types) = 0
```
??

joshuashaffer: It's not used consistently??

```sql 
D select DISTINCT function_name, parameter_types, varargs from duckdb_functions() where parameter_types[1] is null;
```

|           function_name            |  parameter_types   | varargs |
|------------------------------------|--------------------|---------|
| map_from_entries                   | []                 | ANY     |
| list_median                        | [NULL]             |         |
| ... | ... | ... |

Sometimes, they have a type in the Vargas columns; other times, they have an explicit `NULL` as the first argument.

joshuashaffer: @Mause is this ok solution given the above, given that vargargs aren't handled consistently?

Mause: Sounds good for now yeah

Mause (approved):

Mause:
Thanks!

PR #30, opened 2024-06-10
feat: struct, list, and map parameter support 

Mause:


rayokota:
@Mause , any idea on when this will be released?

Mause:
> @Mause , any idea on when this will be released?

It will be included in the next DuckDB release

StephenFlavin:
Hi @Mause,

I was excited to see support for `Connection#createArrayOf` was added so I decided to build locally to try it out and I came across one issue and a question on performance.

## Issue:

```java
connection.createArrayOf("timestamp", new Instant[] {Instant.now()} )
```
produces an error
```
...
Caused by: java.sql.SQLException: Invalid Input Error: Unsupported parameter type
	at org.duckdb.DuckDBNative.duckdb_jdbc_execute(Native Method)
	at org.duckdb.DuckDBPreparedStatement.execute(DuckDBPreparedStatement.java:142)
	at org.duckdb.DuckDBPreparedStatement.executeQuery(DuckDBPreparedStatement.java:174)
Caused by: java.sql.SQLException: Invalid Input Error: Unsupported parameter type
```
Not sure if this is intentional for now or an oversight

## Performance Question:

Before I saw that this had been implemented I passed arrays as literals from java then casting the parameter to the array type in the sql, it seems passing the array literal is twice as fast as using `Connection#createArrayOf`, I tried similar on postgres in the past and the speed improvement using `Connection#createArrayOf` was significant.
Quick test code (using spring)
```java
public static void main(String[] args) throws SQLException {
        var jdbcClient = JdbcClient.create(new SingleConnectionDataSource("jdbc:duckdb:", true));

        System.out.println("ArrayLiteral");

        for (int i = 0; i < 100; i++) {
            final var now2 = Instant.now();
            jdbcClient.sql("select 'hello' from unnest(:arr::int[])")
                    .param("arr", IntStream.range(0, 10000).mapToObj(String::valueOf).collect(Collectors.joining(",", "[", "]")))
                    .query(String.class)
                    .stream().count();
            System.out.println(Duration.between(now2, Instant.now()).toMillis());
        }

        System.out.println("`Connection#createArrayOf`");

        try (final var connection = new SingleConnectionDataSource("jdbc:duckdb:", true).getConnection()) {
            for (int i = 0; i < 100; i++) {
                final var now = Instant.now();
                jdbcClient.sql("select 'hello' from unnest(:arr)")
                        .param("arr", connection.createArrayOf("int", IntStream.range(0, 10000).boxed().toArray()))
                        .query(String.class)
                        .stream().count();
                System.out.println(Duration.between(now, Instant.now()).toMillis());
            }
        }
    }
```
output on M1 mac showing 7ms for array literal vs 13ms for `Connection#createArrayOf`

PR #43, opened 2024-06-24
Make result set getColumn by name case insensitive.  Add test for same.

### LINKED ISSUE #40: Add case-insensitive getters on resultSet
I believe the getters for resultSet should be case insensitive:

>Column names used as input to getter methods are case insensitive. 

From [here](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html)

I'm having an issue where I'm adding duckdb support to an application that already supports a number of other database systems through jdbc. There the results are always fetched like:

```java
resultSet.getString('VERSION') 
```
for a column `version`. This currently doesn't work in duckdb.

jghoman:
Fixes #40 .  Adds unit test both for column names from resultset as well as tablename itself, since we're already there.

This uses `equalsIgnoreCase` rather than the previous `contentEquals,' which is ever so slightly different (String to String versus String to CharSequence/String/StringBuffer, respectively), but I think that shouldn't make a difference in these use cases.

### DIFF
--- a/src/main/java/org/duckdb/DuckDBResultSet.java
+++ b/src/main/java/org/duckdb/DuckDBResultSet.java
@@ -311,7 +311,7 @@ public class DuckDBResultSet implements ResultSet {
             throw new SQLException("ResultSet was closed");
         }
         for (int col_idx = 0; col_idx < meta.column_count; col_idx++) {
-            if (meta.column_names[col_idx].contentEquals(columnLabel)) {
+            if (meta.column_names[col_idx].equalsIgnoreCase(columnLabel)) {
                 return col_idx + 1;
             }
         }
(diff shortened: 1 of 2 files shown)


Mause (changes requested):
Looks great! 

While you're at it, would you be able to fix the other findColumn in the other class?

edigonzales:
Hi @Mause, since i run into this bug: which other class did you mean?

staticlibs:
@jghoman 

Thanks for the PR! I've included it as-is into #185 also adding there the requested changes to the `Array` result set. Including you as a co-author to that PR.

@elefeint 

I wonder if you would like to take a look at the [changes to DuckDBArrayResultSet](https://github.com/duckdb/duckdb-java/pull/185/files#diff-4a34100c3e89e5ed77a395a0216dc09fde76ecf486c1e5ade3d168e7ce379201) (as the original author of this bit)? The changes there are mostly mechanical, but one change of `getLong` to not call `getInt` actually changes the possible output for long values.

staticlibs:
The changes from this PR were included into #185 PR, closing this PR as superseded.

PR #50, opened 2024-07-02
Used a LinkedHashMap for DuckDBStruct.getMap

rayokota:
This retains the order of the attributes in the struct.


rayokota:
@Mause , would it be possible to get a review for this small improvement?  Thanks in advance!

Mause:
Thanks!

PR #57, opened 2024-07-15
Explicitly exit script, avoiding to fail auth, to be reworked

carlopi:


carlopi:
Not really a problem, closing.

PR #60, opened 2024-07-17
DuckDBResultSet bug fix

Stongtong:
1. compatible type is LocalDate;
2. mybatis with duckdb 1.0.0 call getObject throw Exception, impl it

### DIFF
--- a/src/main/java/org/duckdb/DuckDBResultSet.java
+++ b/src/main/java/org/duckdb/DuckDBResultSet.java
@@ -27,6 +27,7 @@ import java.sql.Statement;
 import java.sql.Struct;
 import java.sql.Time;
 import java.sql.Timestamp;
+import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.OffsetTime;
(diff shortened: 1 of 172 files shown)


src/main/java/org/duckdb/DuckDBResultSet.java
```diff
@@ -1286,7 +1295,19 @@ public <T> T getObject(int columnIndex, Class<T> type) throws SQLException {
...
     public <T> T getObject(String columnLabel, Class<T> type) throws SQLException {
-        throw new SQLFeatureNotSupportedException("getObject");
+        if (type == null) {
+            throw new SQLException("type is null");
+        }
+        if (columnLabel == null || columnLabel.isEmpty()) {
+            throw new SQLException("columnLabel is null");
+        }
+        int index = -1;
+        for (int i = 0; i < meta.column_names.length; i++) {
```
Mause, line 1305: Please use the existing findColumn method here instead

Mause (changes requested):
Please add some tests

src/main/java/org/duckdb/DuckDBResultSet.java
Stongtong: add some unit test in TestDuckDBJDBC and use findColumn

Mause:
Thanks!

Issue #66, opened 2024-08-23
Test failure: test_all_types fails because the correct_answer_map does not have any values for the varint column


rpbouman:
Running `make test` fails with:

```
TestDuckDBJDBC#test_all_types failed with java.lang.NullPointerException
java.lang.NullPointerException
        at org.duckdb.TestDuckDBJDBC.test_all_types(TestDuckDBJDBC.java:3981)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.duckdb.test.Runner.runTests(Runner.java:45)
        at org.duckdb.TestDuckDBJDBC.main(TestDuckDBJDBC.java:4525)
```

The test fails unexpectedly because the `correct_answer_map` is not populated with the values for the `varint` column (https://github.com/duckdb/duckdb-java/blob/main/src/test/java/org/duckdb/TestDuckDBJDBC.java#L3854).
 
Apart from populating the map with the types, It would also be nice to assert that the map contains values for the specified column and then report the column name if they are missing.

PR #67, opened 2024-08-23
Fix https://github.com/duckdb/duckdb-java/issues/66

rpbouman:
I wanted to check something around Blob handling and checked out the repo, built it, ran the tests and found two tests failing (TestDuckDBJDBC#test_all_types and TestDuckDBJDBC#test_struct_with_bad_type)

This PR fixes those tests. 

For TestDuckDBJDBC#test_all_types:
- This adds a check to see if the correct_answer_map has values for the column so that the column name can be reported in case the map is not populated for every column from test_all_types.
- It also populates the map with the required values to allow the test to pass.

For TestDuckDBJDBC#test_struct_with_bad_type:
- the assert tried to test by matching the error message with a wrong error message. Or rather, most likely the error message that was tested used to be returned by DuckDB but has changed in the mean while.


rpbouman:
I'm sorry to PR test failed. I tried to figure it out but I believe the compliance action is broken, as it's complaining that it cannot find classes from the test harness itself.

Mause:
> I'm sorry to PR test failed. I tried to figure it out but I believe the compliance action is broken, as it's complaining that it cannot find classes from the test harness itself.

Yeah it's broken at the moment, need to get around to fixing it

Cheers for the other fixes though!

PR #86, opened 2024-09-17
Update jdbc_maven_deploy.py

hannes:


hannes:
this seems to fix the combine issue

PR #88, opened 2024-09-23
Fix metadata reflection for DECIMAL columns - with test

lhofhansl:
JDBC describes that COLUMN_SIZE for a DECIMAL type should reflect its total number of digits, whereas DECIMAL_DIGITS describes it numeric scale.

See: https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)

This does the right thing for all other types (where numeric_precision is null), *except* for INTEGER and LONG, where currently in DuckDB this reports 32 and 64, resp, where it should report 10 and 19. That's for another PR, I think.


Mause:
Thanks!

PR #89, opened 2024-09-23
Allow escaping in schema reflection - with test

lhofhansl:
This allows using \ to escape the LIKE expressions in schema reflection.


Mause:
Thanks!

PR #95, opened 2024-10-07
Use a single transaction per batch

lhofhansl:
I realize that one use the Appender instead, but I did notice that we force a transaction per row even when auto-commit is off.

This uses a transaction per batch instead. Looks like this is already covered in the tests, please let me know if you think that's not sufficient.

Roughly a 2x improvement in tests (this is a using DuckDB from Trino).


hannes:
thanks!

PR #106, opened 2024-12-05
Fix CI

lnkuiper:
Increase `manylinux` and `ubuntu` versions, add `'core_functions'`, and fix compile/test issues.
Looks like a big diff since I vendored in the `'core_functions'`, but I did not change anything else.

CI was successful on my fork:
https://github.com/lnkuiper/duckdb-java/actions/runs/12180166243


carlopi:
Thanks, I think it makes sense.

Curiosity: which DuckDB commit is this connected to? (that is, are extensions needed on that commit? and if so, are they there?)

lnkuiper:
It's connected to the last nightly in the main repo

Mytherin:
Thanks!

PR #115, opened 2024-12-29
Expose db memory address and path to libduckdb_java.so V2

Jens-H:
This PR replaces #87!

Same idea, but:
- using Java long instead of String to hold the memory address
- based on top of main HEAD

I read that DuckDB uses pointer swizzling, which should break before there is problem with the uintptr_t -> jlong cast. This means it should be OK to use it on all supported platforms.

### DIFF
--- a/src/jni/duckdb_java.cpp
+++ b/src/jni/duckdb_java.cpp
@@ -397,6 +397,12 @@ struct ConnectionHolder {
 	}
 };
 
+// Returns the pointer's memory address of duckdb::DuckDB as a jlong
+jlong _duckdb_jdbc_db_memory_address(JNIEnv *env, jclass, jobject conn_ref_buf) {
+	auto conn_ref = (ConnectionHolder *)env->GetDirectBufferAddress(conn_ref_buf);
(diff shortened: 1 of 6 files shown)


Jens-H:
I ran make format and committed some unrelated files and lines with https://github.com/duckdb/duckdb-java/pull/115/commits/dc0833aeb3ad6a211e58b7d38aefbc4968b8eb14 ... seems like without much success.

Jens-H:
To add a rational for this change:
Since Java 22 the JEP 454 "Foreign Function & Memory API" left the preview status (https://openjdk.org/jeps/454) and is now the second official way to interact with native code. 
This would open the option to use the faster C API (like columnar access) instead of the traditional JDBC methods. As JDBC is everywhere it should be possible to have both. Unfortunately the Duckdb database pointer is not accessible outside JNI. With the new method it is possible to rebuild the pointer when needed.

@Mause You reviewed the old PR, so maybe you can have a look.

staticlibs:
Closing this as outdated. If this functionality is still required - this can be revisited.

PR #120, opened 2025-01-07
Build duckdb_java linux-amd64-musl

hrl20:
Build duckdb_java for Alpine linux on amd64

(arm64 is possible but takes [4+ hours](https://github.com/hrl20/duckdb-java/actions/runs/12589697140) on the default GHA runner, so skipping for now)


.github/workflows/Java.yml
```diff
@@ -196,6 +196,58 @@ jobs:
...
             build/release/duckdb_jdbc.jar
 
+  java-linux-amd64-musl:
+    name: Java Musllinux (amd64)
+    runs-on: ubuntu-latest
+    container:
+      image: alpine:latest 
+    env:
+      GEN: ninja
+      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
```
carlopi, line 206:
```suggestion
```
I think this is no more supported / not useful

CMakeLists.txt
```diff
@@ -65,12 +64,31 @@ if(OVERRIDE_JDBC_OS_ARCH)
...
+        set(${RESULT} FALSE PARENT_SCOPE)
+      endif()
+    else()
+      set(${RESULT} FALSE PARENT_SCOPE)
+    endif()
+  else()
+    set(${RESULT} FALSE PARENT_SCOPE)
+  endif()
+endfunction()
+
```
carlopi, line 86: Nice! Would you mind sending a PR with this also to duckdb/duckdb ?
Current detection is somehow worse, I think this might be handy in general.

carlopi (commented):
Main question: can we get away from static linking the libc++ ?
I see that working with, question is if we can manage without.

.github/workflows/Java.yml
```diff
@@ -107,7 +107,7 @@ jobs:
          # ./scripts/upload-assets-to-staging.sh github_release duckdb_jdbc-linux-aarch64.jar
 
       - uses: actions/upload-artifact@v3
-        with:
+        with: duckdb_jdbc-linux-aarch64
```
Mause, line 110: this doesn't look valid, can you please remove it?

hrl20:
> Main question: can we get away from static linking the libc++ ? I see that working with, question is if we can manage without.

I think in general this will not work because the standard alpine distribution doesn't include these libraries. Not linking it means requiring users to always install them in their Alpine containers when they want to use duckdb. This error happened trying to run duckdb (libstdc++ and libgcc linked) in the standard Metabase container which uses alpine, and loading httpfs (no libstdc++ and libgcc linked). 

Screenshot 2024-10-29 at 1 43 20 AM (1)


Also this does mean the extensions need to have statically linked libstdc++ and libgcc. Idk if there's a better solution here.

.github/workflows/Java.yml
hrl20: not sure how this got there. thank you!

CMakeLists.txt
hrl20: sure.

carlopi:
But I think the problem here is that v1.1.2 has no concept (yet) of musl.

Would it be possible to run the same workflow again using a recent duckdb commit (such as 09cef57892 the one currently backed in duckdb-java)

hrl20:
> But I think the problem here is that v1.1.2 has no concept (yet) of musl.
> 
> Would it be possible to run the same workflow again using a recent duckdb commit (such as 09cef57892 the one currently backed in duckdb-java)

I tried this again on [this branch](https://github.com/hrl20/duckdb-java/tree/refs/heads/build-linux-amd64-only-musl-nolibcpp) and it still fails `/tmp/libduckdb_java14876668067405368824.so: Error loading shared library libstdc++.so.6: No such file or directory (needed by /tmp/libduckdb_java14876668067405368824.so)`

staubt:
Hello
Is it planned to bring that to one of the next releases?
I would like to use duckdb java with an alpine base image.

staticlibs:
@staubt 

Hi, yes, preliminary JDBC `v1.3.0.x` is going to include the build with musl libc. Though it was found that on some workloads DuckDB on musl may be multiple times slower than DuckDB on glibc, so should be used with caution.

Also, may I ask which JDK build are you considering to use on Alpine, is it glibc-based or musl-based?

staubt:
Hi
Thanks for the fast answer. The Alpine image I want to use is musl-based.

carlopi:
@hrl20: I realize my previous comments are void, thanks for the explanation.

carlopi:
@staticlibs: unsure how to proceed, I think it's more appropriate for you to make a call

PR #123, opened 2025-01-08
Redirect prepareStatment

### LINKED ISSUE #82: DuckDBConnection.prepareStatement methods could be less strict
The `prepareStatement` methods in `DuckDBConnection` here:
https://github.com/duckdb/duckdb-java/blob/f7ae8cb492d5ffedec1042fbcfacc14ea3e03de5/src/main/java/org/duckdb/DuckDBConnection.java#L296

They are currently all stubbed with `throw new SQLFeatureNotSupportedException("prepareStatement")`. 

Untill they are fully implemented, they could be rewritten so that certain calls may proceed if generated keys are not required. For instance, the call at L296 could be rewritten to:
```java
public PreparedStatement prepareStatement(String sql, int autoGeneratedKeys) throws SQLException {
    if (autoGeneratedKeys == NO_GENERATED_KEYS) {
        return prepareStatement(sql);
    } else {
        throw new SQLFeatureNotSupportedException("prepareStatement");
    }
}
```

These methods as written are currently keeping me from utilizing duckdb with the [Jetbrains Exposed](https://github.com/JetBrains/Exposed) API, because the connection wrappers call `prepareStatement([sql], NO_GENERATED_KEYS)` instead of `prepareStatement([sql])`.

adiykehler:
I tried to implement the proposed change. Closes #82 .


rmannibucau:
maybe join https://github.com/duckdb/duckdb-java/pull/829 since i'm trying to not silently swallow unsupported parameters

staticlibs:
Closing as superseded by #829

PR #132, opened 2025-01-21
JDBC Timestamps

laurentco:
See https://github.com/duckdb/duckdb/pull/10736


laurentco:
Closing this PR after reading more on the subject of storing dates with TZ, I was under the impression that `setTimestampTZ` of similar was missing in hte API but now realise TZ is not stored in Duckdb including when using `TIMESTAMP WITH TZ` type and this is compatible with postgres.
Storing all dates as UTC then makes sense, hence having only `setOffsetDateTime` with TZ ignored and hardcoded to UTC is sufficient.

PR #131, opened 2025-01-19
Bundle jemalloc with JDBC Driver

brianwyka:
Fixes #63

PR #133, opened 2025-01-22
Bundle jemalloc

lnkuiper:
Trying to help out #131 

I don't think it's necessary to vendor in the changes, the CI will do that by itself. Just changing the vendor script should be enough. This should hopefully avoid conflicts.

Big thanks to @brianwyka for making the changes to the script! I simply copied them here.


lnkuiper:
I didn't touch the file that failed the format check, so I think it's fine

brianwyka:
Thanks @lnkuiper !

lnkuiper:
@hannes, I've fixed the format check. The issue was not wrongly formatted code, but the clang-format version in CI was wrong. I've now made it install the same version as we use in the main duckdb repo again.

brianwyka:
Thanks for getting this merged!

PR #135, opened 2025-01-25
Get profiling information api

akoshchiy:
Added `DuckDBConnection#getProfilingInformation`, which wraps `Connection::GetProfilingInformation` c++ api.


akoshchiy:
Seems, that format check is failing on the `main`.

Issue #143, opened 2025-02-20
JDBC driver built with JDK 11 cannot be used with JDK 8


staticlibs:
When building the Java part, `-source` and `-target` options [are set to 1.8](https://github.com/duckdb/duckdb-java/blob/bb7eefe73ff4783806943ae228a9b25971a5fea2/CMakeLists.txt#L30), but `-bootclasspath` option is not set, so it gives the following warning during compilation:

```
warning: [options] bootstrap class path not set in conjunction with -source 8
```

And fails the following way when run on Java 8:

```
Exception in thread "main" java.lang.NoSuchMethodError: java.nio.ByteBuffer.position(I)Ljava/nio/ByteBuffer;
    at org.duckdb.DuckDBResultSet$DuckDBBlobResult.<init>(DuckDBResultSet.java:435)
    at org.duckdb.TestDuckDBJDBC.blobOf(TestDuckDBJDBC.java:3954)
    at org.duckdb.TestDuckDBJDBC.<clinit>(TestDuckDBJDBC.java:4045)
```

PR #144, opened 2025-02-20
Remove java cross-comp options, include debuginfo

staticlibs:
When compiling Java part with JDK 11 (or later), the current javac options `-source 1.8` and `-tatget 1.8` are ineffective. Resulting bytecode format is compatible with JDK 8, but cannot be run on JDK 8 because it is compiled against the bootstrap classpath of JDK 11 and depends on methods not available in JDK 8.

This change removes these cross-compilation flags. It is assumed, that official release binaries are always built with JDK 8. For users/packagers, who is going to build the driver from source, a CMake configure-time warning about JDK 8 compatibility is added.

Additionally Java debuginfo is added to the JAR (`-g` flag). It does not affect the performance and the JAR size increase is minor (less than 10KB).

Fixes: #143


hannes (approved):
Great solution, did not know this was an issue

staticlibs:
Thanks for the review! @Mytherin mentioned that he can merge this PR (and #164 along with it) after they have completed the CI runs after the rebase.

Mytherin:
Thanks!

Issue #137, opened 2025-02-11
[v1.2.0] Failed to set configuration option "allowed_directories", error: Attempted to dereference unique_ptr that is NULL!


jivanic-demystdata:
When trying to set the `allowed_directories` configuration when starting a new DuckDB instance, I have the following error:

```
Exception in thread "main" java.sql.SQLException: Catalog Error: Failed to set configuration option "allowed_directories", error: Attempted to dereference unique_ptr that is NULL!
        at org.duckdb.DuckDBNative.duckdb_jdbc_startup(Native Method)
        at org.duckdb.DuckDBConnection.newConnection(DuckDBConnection.java:51)
        at org.duckdb.DuckDBDriver.connect(DuckDBDriver.java:41)
        at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:683)
        at java.sql/java.sql.DriverManager.getConnection(DriverManager.java:191)
        at duck$_.<init>(duck.sc:14)
        at duck_sc$.script$lzyINIT1(duck.sc:29)
        at duck_sc$.script(duck.sc:29)
        at duck_sc$.main(duck.sc:33)
        at duck_sc.main(duck.sc)
```

To reproduce:
1. Install `scala-cli`
    See https://scala-cli.virtuslab.org/install
2. Create a new file named `duck.sc`
3. Copy/paste the following code into the `duck.sc` file:
```scala
//> using dep org.duckdb:duckdb_jdbc:1.2.0

import java.sql.DriverManager
import java.util.Properties

val config = new Properties()
config.put("enable_external_access", "false")
// allowed_directories value took from https://github.com/duckdb/duckdb/blob/f317a112e0c789a11987d9de6d621d8469c00fb8/test/sql/settings/allowed_directories.test#L16
config.put("allowed_directories", "['data/csv/glob']")

DriverManager.getConnection("jdbc:duckdb:", config)
```
4. Run the `duck.sc` file with the bash cmd: 
```bash
scala-cli duck.sc
```

Notice that commenting the line `config.put("allowed_directories", "['data/csv/glob']")` removes the error

staticlibs:
@jivanic-demystdata Hi! Will look at this shortly.

staticlibs:
[...]

PR #159, opened 2025-03-13
Support typed configuration options

staticlibs:
This change adds support for all primitive types (and additionally for a list of strings) for configuration values that can be passed to `DriverManager.getConnection(url, conf)`.

To be able to handle config in a separate compilation unit, cached references and common utilities are moved to separate files. There are a few more check added to the Java references collection code, but it still kept largely the same.

Additionally the `file_system` field of the `DBConfig` is filled manually, it was necessary because this field is used for checks when incoming file-system related options are set.

Testing: new test added that sets typed config values and checks that they are set with `current_setting()`.

Fixes: #137


Mytherin:
Thanks!

Issue #147, opened 2024-02-29
Mismatch between uuid bind to as variable vs value returned from db with jdbc


solita-kattelus:
### What happens?

Testing new UUID binding and somehow get mismatching UUID back
```
    @Test
    public void testUUIDBinding() throws SQLException {
        UUID uuid = UUID.fromString("0b17ce61-375c-4ad8-97b3-349d96d35ab1");
        Connection conn = DriverManager.getConnection("jdbc:duckdb:");
        PreparedStatement statement = conn.prepareStatement("select ? as value");
        statement.setObject(1, uuid);
        ResultSet resultSet = statement.executeQuery();
        Assertions.assertTrue(resultSet.next());
        Object object = resultSet.getObject(1);
        Assertions.assertEquals(uuid, object);
        Assertions.assertFalse(resultSet.next());
    }
```

And it fails with Assertions.assertEquals(uuid, object); check as:
```
Expected :0b17ce61-375c-4ad8-97b3-349d96d35ab1
Actual   :8b17ce61-375c-4ad8-97b3-349d96d35ab1
```

### To Reproduce

Reproduce can be done with junit test above.

### OS:

macOs arm64

### DuckDB Version:

0.10.0

### DuckDB Client:

JDBC

### Full Name:

Marko Kattelus

### Affiliation:

Solita Oy

### Have you tried this on the latest [nightly build](https://duckdb.org/docs/installation/?version=main)?

I have tested with a release build (and could not test with a nightly build)

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [X] Yes, I have

guizmaii:
@staticlibs Are you guys publishing snapshot versions? Not sure we wanna wait for `v1.2.2`, scheduled in April, for having access to this fix

staticlibs:
@guizmaii 

Nightly builds (that are expected to be done from the latest `main`) are available [here](https://duckdb.org/docs/installation/?version=main&environment=java&download_method=direct), but the PR needs to be merged first to get there. I will ask around tomorrow if it is possible to expedite JDBC PRs reviews.

PR #160, opened 2025-03-14
Modify UUID when passing it as a bind variable

staticlibs:
When parsing an UUID from a string literal DuckDB modifies it with [flipping the most significant bit](https://github.com/duckdb/duckdb/blob/c9dd40f144e07a5dddb1d81c7218555205894c80/src/common/types/uuid.cpp#L51) in `UUID::FromString` for the following reason:

"Flip the first bit to make `order by uuid` same as `order by uuid::varchar`"

When an UUID is read from a result set, JDBC driver flips the MSB again to get back the original UUID in Java.

This bit-flipping logic was missing when an UUID was passed from Java as a bind variable. The proposed change adds the same logic when passing bind variables.

Testing: existing test is improved to cover both string literal and a bind variable.

Fixes: #147

### DIFF
--- a/src/jni/duckdb_java.cpp
+++ b/src/jni/duckdb_java.cpp
@@ -13,6 +13,9 @@
 #include "duckdb/parser/parsed_data/create_type_info.hpp"
 #include "functions.hpp"
 
+#include <cstdint>
+#include <limits>
+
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #162, opened 2025-03-14
Do not build jemalloc on Windows

staticlibs:
The jemalloc extention was recently added to JDBC builds in #133. The logic for its selection in `vendor.py` does not look fully correct to me, but both Linux and Mac are passing on CI, only Windows is breaking, so this change only tries to unbreak Windows builds, and the jemalloc selection logic can be improved in subsequent PRs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,74 @@ set(CMAKE_JAVA_COMPILE_FLAGS -source 1.8 -target 1.8 -encoding utf-8)
 add_definitions(-DDUCKDB_BUILD_LIBRARY)
 
 if(MSVC)
+  remove_definitions(-DDUCKDB_EXTENSION_JEMALLOC_LINKED)
+  list(REMOVE_ITEM DUCKDB_SRC_FILES
+    src/duckdb/extension/jemalloc/jemalloc_extension.cpp
+    src/duckdb/extension/jemalloc/jemalloc/src/jemalloc.c
+    src/duckdb/extension/jemalloc/jemalloc/src/arena.c
+    src/duckdb/extension/jemalloc/jemalloc/src/background_thread.c
+    src/duckdb/extension/jemalloc/jemalloc/src/base.c
+    src/duckdb/extension/jemalloc/jemalloc/src/batcher.c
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -32,6 +32,74 @@ set(CMAKE_JAVA_COMPILE_FLAGS -source 1.8 -target 1.8 -encoding utf-8)
 add_definitions(-DDUCKDB_BUILD_LIBRARY)
 
 if(MSVC)
+  remove_definitions(-DDUCKDB_EXTENSION_JEMALLOC_LINKED)
```
Mytherin, line 35: Instead of removing them can we perhaps not add them in the first place? That seems like it would be cleaner. Or is the above list of files auto-generated?

Mytherin (commented):
Thanks for the PR! Looks good - one comment

Mytherin:
We can clean up the logic in a subsequent PR - thanks for the fix!

CMakeLists.txt
staticlibs: Yes, these definitions and the sources list is auto-generated by `vendor.py`, regarding jemalloc it contains the [following logic](https://github.com/duckdb/duckdb-java/blob/a179e869e51f43fda865cc37459cae8dad533fe2/vendor.py#L25). I cannot make sense of it, because these platform checks are done when the `vendor.py` itself is being run (how is it supposed to check for Android and WASM at that time, why checking for WASM for JDBC sources etc - it was added in #133, I asked Laurens in Discord about it). I assume these checks for jemalloc need to be run at CMake configure time instead. And the list of sources needs to be added by `vendor.py` into a separate CMake variable. But for now, without touching `vendor.py` this "filter out on Windows" logic looks the easiest.

staticlibs:
Thanks for merging this! CI is passing now in main, I've rebased other PRs onto the latest main.

CMakeLists.txt
Mytherin: That makes sense. It looks like these checks were copied from the Python build, which indeed doesn't work because the vendor script does not get run per configuration. The correct checks should be in the CMake. I would copy over the checks [from here](https://github.com/duckdb/duckdb/blob/main/extension/extension_config.cmake#L16).

PR #166, opened 2025-03-19
Result set Timestamp handling improvements

staticlibs:
DuckDB supports both `TIMESTAMP WITHOUT TIME ZONE` and `TIMESTAMP WITH TIME ZONE`. When gettting Java `Timestamp` from a result set, user can optionally request desired time zone by specifying a `Calendar` instance.

Thus we are getting the following invariants:

1. `TIMESTAMP WITH TIME ZONE`, no `Calendar`: returns `Timestamp` in a system-default (JVM) time zone that represents the same moment in time as the DB's `TIMESTAMP WITH TIME ZONE`. For example with a system-default time zone being `EET` (`UTC+2`) the timestamp with `12:00` time and `CET` (`UTC+1`) DB time zone will be returned with `13:00` time (in `EET`).

2. `TIMESTAMP WITH TIME ZONE`, user-specified Calendar: the same logic is used as in point `1.`, just the time-zone from the user-specified Calendar is used instead of the system-default time zone.

3. `TIMESTAMP WITHOUT TIME ZONE`, no `Calendar`: returns the timestamp that has the same numeric values of all components as the DB's `TIMESTAMP WITHOUT TIME ZONE`. For example the timestamp with `12:00` time will be returned with the same `12:00` time in any system-defaut time zone.

4. `TIMESTAMP WITHOUT TIME ZONE`, user-specified `Calendar`: this case is tricky because user is requesting to adjust the timestamp for the specified timezone, but DB's timestamp does not denote any time zone that can be used as a starting point for adjustment. In this case we are using the same logic as Postgres' JDBC driver:

https://github.com/pgjdbc/pgjdbc/blob/d86d7ce19030230f14e573b8095f04721d1d8e35/pgjdbc/src/main/java/org/postgresql/jdbc/TimestampUtils.java#L1606

So we assume that the incoming time stamp belongs to the system-default time zone. Then we calculating the offset between the system-default and the user-requested time zones that is effective at the moment of time defined by our timestamp in a system-default time zone. Note, that such offset can be different depending on the time of year due to DST. And then shifting the timestamp by this offset - this can seen as a "reverse" shift comparing to points `1.` and `2.`. For example, with a system-default time zone being `EET` (`UTC+2`) and user-requested time zone being `CET` (`UTC+1`), the timestamp with time `12:00` will be shifted by the difference between `EET` and `CET` and returned as `13:00` (in `CET`).

Additionally the logic added for getting date-only and time-only values from the `TIMESTAMP_*` fields is added. It is done by getting the Java `Timestamp` first (using the invariants described above) and then getting date-only or time-only part of this timestamp.

Testing: existing tests adjusted to match the new time zone handling logic; new test is added for getting `Date` and `Time` from `TIMESTAMP`.

Fixes: #146

### DIFF
--- a/src/main/java/org/duckdb/DuckDBResultSet.java
+++ b/src/main/java/org/duckdb/DuckDBResultSet.java
@@ -31,11 +31,7 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.OffsetTime;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Map;
-import java.util.Objects;
-import java.util.UUID;
+import java.util.*;
 
 public class DuckDBResultSet implements ResultSet {
     private final DuckDBPreparedStatement stmt;
@@ -372,7 +368,7 @@ public class DuckDBResultSet implements ResultSet {
     }
 
     public Time getTime(int columnIndex) throws SQLException {
-        return check_and_null(columnIndex) ? null : current_chunk[columnIndex - 1].getTime(chunk_idx - 1);
+        return getTime(columnIndex, null);
     }
 
     public Timestamp getTimestamp(int columnIndex) throws SQLException {
@@ -911,7 +907,10 @@ public class DuckDBResultSet implements ResultSet {
     }
 
     public Time getTime(int columnIndex, Calendar cal) throws SQLException {
-        return getTime(columnIndex);
+        if (check_and_null(columnIndex)) {
+            return null;
+        }
+        return current_chunk[columnIndex - 1].getTime(chunk_idx - 1, cal);
     }
 
     public Time getTime(String columnLabel, Calendar cal) throws SQLException {
--- a/src/main/java/org/duckdb/DuckDBTimestamp.java
+++ b/src/main/java/org/duckdb/DuckDBTimestamp.java
@@ -1,15 +1,10 @@
 package org.duckdb;
 
 import java.sql.Date;
+import java.sql.SQLException;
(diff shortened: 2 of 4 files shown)


Mytherin:
Thanks for the PR! Matching the logic of the Postgres driver makes sense to me. 

> TIMESTAMP WITH TIME ZONE, no Calendar: returns Timestamp in a system-default (JVM) time zone that represents the same moment in time as the DB's TIMESTAMP WITH TIME ZONE. For example with a system-default time zone being EET (UTC+2) the timestamp with 12:00 time and CET (UTC+1) DB time zone will be returned with 13:00 time (in EET).

This part is a bit unclear to me - `TIMESTAMP WITH TIME ZONE` is internally stored in DuckDB as a `UTC` offset - so returning that offset always returns a `UTC` time regardless of the user-specified timezone in the database (unless a string conversion is triggered inside DuckDB - which bins the timestamp to the user-specified timezone).

I think that behavior is fine and makes sense - but just figured I would mention this in case that was not clear.

staticlibs:
Thanks for the review! Yes, the wording here is ambiguous, under "DB time zone" I mean the time zone offset that is already included with the DB's `TIMESTAMP WITH TIMEZONE`  value, NOT the DB-level default time zone (e.g. set by user with `SET TIMEZONE`). In this case the timestamp value fetched from DB (that already includes the offset in its UTC value) is adjusted to the JVM-default time zone. DB-level default time zone does NOT affect the conversion in any of 4 cases.

Mytherin:
Alright - thanks for the clarification!

Issue #165, opened 2025-03-18
`DatabaseMetaData.getFunctions(…)` should accept a null schema


desruisseaux:
The [JDBC Javadoc](https://docs.oracle.com/en/java/javase/23/docs/api/java.sql/java/sql/DatabaseMetaData.html#getFunctions(java.lang.String,java.lang.String,java.lang.String)) of `DatabaseMetaData.getFunctions(…)`  said:

> `schemaPattern` - a schema name pattern; must match the schema name as it is stored in the database; "" retrieves those without a schema; null means that the schema name should not be used to narrow the search

Therefore, setting the `schemaPattern` argument to `null` in the following code snippet should return all functions. However, that code prints nothing with DuckDB 1.2.1:

```java
import java.sql.Connection;
import java.sql.DriverManager;
import java.sql.ResultSet;

public class TestDuckDB {
    public static void main(String[] args) throws Exception {
        try (Connection c = DriverManager.getConnection("jdbc:duckdb:/...something.db")) {
            try (ResultSet r = c.getMetaData().getFunctions(null, null, "%")) {
                while (r.next()) {
                    System.out.println(r.getString("FUNCTION_NAME"));
                }
            }
        }
    }
}
```

If we replace the `schemaPattern` argument (the last `null`) by `"%"` in the call to `getFunctions(…)`, it works. The fix for this issue may be simply to replace `null` by `"%"` internally in the JDBC driver.

desruisseaux:
It also seems that `toFunctionNames(…)` does not recognize the escape character declared in `getSearchStringEscape()`, which is `\`. When the spatial extension is installed (`LOAD spatial`) , the following gives me the list of spatial functions as expected:

```java
c.getMetaData().getFunctions(null, "%", "ST_%")
```

But if I replace `"ST_%"` by `"ST\\_%"` in order to require specifically the underscore character instead of any character, then I get an empty `ResultSet`.

desruisseaux:
Thanks!

PR #167, opened 2025-03-19
DatabaseMetadata schema introspection enhancements

staticlibs:
A few schema introspection functions in `DatabaseMetadata` take `catalog` and `schemaPattern` parameters with the following specific requirements:

> catalog - a catalog name; must match the catalog name as it is stored in the database; "" retrieves those without a catalog; null means that the catalog name should not be used to narrow the search
> schemaPattern - a schema name pattern; must match the schema name as it is stored in the database; "" retrieves those without a schema; null means that the schema name should not be used to narrow the search

This change makes the handling for these parameters consistent between `getSchemas`, `getTables`, `getColumns`, `getPrimaryKeys`, `getIndexInfo` and `getFunctions` functions. Additionally it fixes the escaping for `functionNamePattern` parameter in `getFunctions`.

Testing: test coverage added for enhanced `catalog` and `schemaPattern` handling.

Fixes: #165

### DIFF
--- a/src/main/java/org/duckdb/DuckDBDatabaseMetaData.java
+++ b/src/main/java/org/duckdb/DuckDBDatabaseMetaData.java
@@ -19,6 +19,8 @@ import javax.sql.rowset.RowSetProvider;
 
 public class DuckDBDatabaseMetaData implements DatabaseMetaData {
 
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #172, opened 2025-03-20
Use engine commit hash when vendoring sources

staticlibs:
This PR brings duckdb/duckdb-odbc#84 changes to JDBC repo.

Testing: no code changes, no new tests.

### DIFF
--- a/.github/workflows/Vendor.yml
+++ b/.github/workflows/Vendor.yml
@@ -1,10 +1,5 @@
+name: Vendor
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - ".github/workflows/Nightly.yml"
-      - "vendor.py"
   workflow_call:
     inputs:
       duckdb-sha:
@@ -19,10 +14,6 @@ on:
         required: false
         default: ''
         type: 'string'
-  schedule:
-    - cron: "17 0 * * *"
-
-name: Vendor upstream DuckDB sources
 
 jobs:
   vendor:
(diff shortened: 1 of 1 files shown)


.github/workflows/Nightly.yml
```diff
@@ -58,6 +59,10 @@ jobs:
         run: |
           git push -u origin HEAD
 
+      - name: Wait for push to complete
+        if: steps.vendor.outputs.vendor != '' && github.event_name != 'pull_request'
+        run: sleep 10  # Give GitHub a moment to process the push
```
Mytherin, line 64: Could you expand on why this sleep is necessary? Doesn't the push complete otherwise?

staticlibs: M, I've just copied this bit from the latest ODBC workflow to keep JDBC in sync with it. I will re-check.

staticlibs: As an update on this, I've got some input on GH workflows from Carlo, as a result the vendoring process in ODBC is [being changed](https://github.com/duckdb/duckdb-odbc/pull/96) to import engine sources into a separate branch, create a PR on this branch and merge this PR automatically. In this variant the testing workflow is dispatched by triggering the `pull_request` event instead of a direct dispatch, so we do not need to wait manually until the push completes. I intend to bring the same workflow changes to JDBC once they are approved for ODBC, keeping this PR as a "draft" for now.

staticlibs: Removed the timeout, now using `push` event instead that is expected only to be triggered by Github after the push is actually complete. Also removed the periodic trigger and submitted duckdb/duckdb/pull#17099 on the main DuckDB repo side to notify the JDBC vendoring process.

staticlibs:
@carlopi I wonder if you can look at this PR too along with the duckdb/duckdb/pull#17099 .

staticlibs:
Merging without review as a workflow-only change, and also to create the GH API entry point on `duckdb-java` side to be ready for upcoming calls from mainline DuckDB repo, that were enabled there earlier today with merged duckdb/duckdb#17099 .

PR #176, opened 2025-03-24
Retain client config on connection duplication

staticlibs:
This change copies `ClientConfig` to a new connection on `duplicate()` call.

Testing: new test added that checks that `home_directory` value is retained in a duplicated connection.

Fixes: #148


Mytherin:
Thanks!

PR #178, opened 2025-03-27
add types.cpp to jni build

w1nk:
when building on alpine (using its toolchain), without this file being explicitly listed, the resulting build fails with symbol errors in the linked .so.    adding this file explicitly to the build resolves them.


staticlibs:
Hi, thanks for the PR! Sorry, I don't have permission to approve the workflow run for this PR.

Note, the same change in included with #170 PR (so either it or this one should be merged to fix the CI). Also, the same change is required in `CMakeLists.txt.in` that is used to overwrite `CMakeLists.txt` when new engine sources are imported.

staticlibs:
Closing this in favor of #187.

Issue #121, opened 2025-01-08
ParameterMetadata is only partially implemented


eramitmittal:
On the lines of pre-existing implementation of corresponding methods for columns in ResultSetMetaData, ParameterMetadata should also have implementations at-least for following methods:

isSigned
getPrecision
getScale
getParameterType
getParameterTypeName
getParameterClassName

Issue #41, opened 2024-06-22
The DuckDBParameterMetaData class is not really implemented


rsmckinney:
The DuckDBParameterMetaData class is not really implemented. This hinders usability and type-safety of code generated for duckdb. For  instance, type information of this sort is vital for the [manifold-sql](https://github.com/manifold-systems/manifold/blob/master/manifold-deps-parent/manifold-sql/readme.md) project. Please consider properly implementing the methods for this class. Thanks.

PR #114, opened 2024-12-20
Adding fuller implementation of ParameterMetadata

eramitmittal:
Fixes #121 
On the lines of pre-existing implementation of corresponding methods for columns in ResultSetMetaData, adding ParameterMetadata implementations for
1. isSigned
2. getPrecision
3.  getScale
4. getParameterType
5. getParameterTypeName
6. getParameterClassName

PR #181, opened 2025-03-31
Implement ParameterMetadata

staticlibs:
This PR brings in the change that was contributed earlier by `eramitmittal` in #114 but was not merged. That change is rebased onto the latest `main`with minimal adjustetments.

Testing: in addition to the `DECIMAL` test from original PR more tests are added to cover all other column types.

Fixes: #121
Fixes: #41


staticlibs:
Just for the record, the compliance/TCK test suite failure is caused by the outdated CMake min version declared in the [test runner](https://github.com/cwida/jdbccts), I've submitted [the PR](https://github.com/cwida/jdbccts/pull/1) to fix that.

Mytherin:
Thanks!

Issue #55, opened 2024-07-12
feat: support spark jdbc write to duckdb.


chaozwn:
spark uses path to store catalog information, which is the option injected by spark by default. duckdb-java is recommended to add the path parameter recognition, at least do not throw RuntimeException.

```
16:28:57    Runtime Error in model movies (models/sources/movies.sql)
  An error occurred while calling o40.sql.
  : java.sql.SQLException: Invalid Input Error: Unrecognized configuration property "path"
        at org.duckdb.DuckDBNative.duckdb_jdbc_startup(Native Method)
        at org.duckdb.DuckDBConnection.newConnection(DuckDBConnection.java:48)
        at org.duckdb.DuckDBDriver.connect(DuckDBDriver.java:41)
        at org.apache.spark.sql.execution.datasources.jdbc.connection.BasicConnectionProvider.getConnection(BasicConnectionProvider.scala:49)
        at org.apache.spark.sql.execution.datasources.jdbc.connection.ConnectionProviderBase.create(ConnectionProvider.scala:102)
        at org.apache.spark.sql.jdbc.JdbcDialect.$anonfun$createConnectionFactory$1(JdbcDialects.scala:123)
        at org.apache.spark.sql.jdbc.JdbcDialect.$anonfun$createConnectionFactory$1$adapted(JdbcDialects.scala:119)
        at org.apache.spark.sql.execution.datasources.jdbc.JdbcRelationProvider.createRelation(JdbcRelationProvider.scala:50)
        at org.apache.spark.sql.execution.datasources.DataSource.writeAndRead(DataSource.scala:506)
        at org.apache.spark.sql.execution.command.CreateDataSourceTableAsSelectCommand.saveDataIntoTable(createDataSourceTables.scala:228)
        at org.apache.spark.sql.execution.command.CreateDataSourceTableAsSelectCommand.run(createDataSourceTables.scala:183)
        at org.apache.spark.sql.execution.command.ExecutedCommandExec.sideEffectResult$lzycompute(commands.scala:75)
        at org.apache.spark.sql.execution.command.ExecutedCommandExec.sideEffectResult(commands.scala:73)
        at org.apache.spark.sql.execution.command.ExecutedCommandExec.executeCollect(commands.scala:84)
        at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.$anonfun$applyOrElse$1(QueryExecution.scala:98)
[...]

PR #184, opened 2025-04-03
Ignore path option passed by Spark

staticlibs:
Apache Spark passes the `path` option when a SELECT on a JDBC DataSource table is performed. It is the internal Spark option and is likely passed by mistake, so we need to ignore it to allow the connection to be established.

Testing: new test added that checks that the `path` option is ignored.

Fixes: #55


staticlibs:
Merging as reviewed offline.

PR #185, opened 2025-04-03
Make ResultSet getters case insensitive

staticlibs:
This change brings in the #43 PR with additional changes to the `Array` result set requested by the reviewer in #43.

For the `Array` result set the column names `INDEX` and `VALUE` are taken from the [implementation in Postgres' pgjdbc driver](https://github.com/pgjdbc/pgjdbc/blob/156d724e1d95052b41a19fb568b2f81919ae2197/pgjdbc/src/main/java/org/postgresql/jdbc/PgArray.java#L399).

Testing: test from #43 is included, test for `Array` result set is enhanced with additional checks.

Fixes: #40

### DIFF
--- a/src/main/java/org/duckdb/DuckDBArrayResultSet.java
+++ b/src/main/java/org/duckdb/DuckDBArrayResultSet.java
@@ -41,22 +41,23 @@ public class DuckDBArrayResultSet implements ResultSet {
         return wasNull;
     }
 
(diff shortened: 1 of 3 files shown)


elefeint (approved):
Thank you! I am pretty sure the original `getInt()` was a mistake.

Mytherin:
Thanks!

PR #187, opened 2025-04-07
Fix builds by adding missing types.cpp to CMake

staticlibs:
In #168 `types.cpp` file was added to `CMakeLists.txt`, but, by mistake it was not added to `CMakeLists.txt.in` that is used to generate the `CMakeLists.txt` file on periodic engine sources import.

This change adds `types.cpp` to both `CMakeLists.txt.in` and `CMakeLists.txt` to fix the build.


staticlibs:
Merging this without review to unbreak the builds.

Issue #191, opened 2025-04-08
Ubuntu Focal 20.04 being removed from GitHub actions


staticlibs:
Latest commits are failing CI runs with:

> This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101

Most GH workflows use the `ubuntu-latest` image, but JDBC compliance tests (run with [these scripts](https://github.com/cwida/jdbccts/)) still run on 20.04. This workflow needs to be updated to newer Ubuntu.

PR #192, opened 2025-04-08
Update Ubuntu image used for compiance runs

staticlibs:
This change attemts to use `ubuntu-latest` instead of `ubuntu-20.04` for JDBC compliance test suite run on GH actions.

Fixes: #191


staticlibs:
Merging without a review as a workflow-only fix.

PR #193, opened 2025-04-10
Enable 4-part version number

staticlibs:
DuckDB is using 3-part version number. Sometimes it may be necessary to make an updated JDBC release with the changes only in JDBC/JNI code keeping the engine version the same. For this it is proposed to add fourth component to version number for JDBC-specific changes. So the version is supposed to look like `v1.2.2.0`.

Testing: JAR bundling with 4-part version was tested in a separate repo.


staticlibs:
Merging without review as a workflow-only change.

PR #214, opened 2025-05-08
Expose query progress on statement

staticlibs:
This change exposes the [duckdb_query_progress](https://duckdb.org/docs/stable/clients/c/api.html#duckdb_query_progress) native call as `getQueryProgress()` method on the statement. Similar to the `Statement#cancel()` call this new method also expected to be called when `execute()` is running and also works on the connection (not on the statement itself).

Note, the local option `SET enable_progress_bar = true` needs to be set to be able to fetch actual progress data.

Testing: new test is added that checks that `getQueryProgress()` returns valid values.


staticlibs:
@lnkuiper I wonder if you would like to take a look at this PR?

lnkuiper (approved):
@staticlibs I have limited JDBC experience, but this looks excellent! Thanks for implementing this so quickly. This is good to go from my side.

staticlibs:
Thanks for the review!

Issue #212, opened 2025-05-06
[Feature Request] Add support to `java.sql.Statement#setQueryTimeout`


jivanic-demystdata:
Right now, the implementation of `java.sql.Statement#setQueryTimeout` in `DuckDBPreparedStatement` is:

<img width="516" alt="Image" src="https://github.com/user-attachments/assets/9f40de47-d28b-47bd-b51b-5f3fc6cb161a" />

See https://github.com/duckdb/duckdb-java/blob/1dfa6c482d1a2566a2f653de93fe365ca8f3d043/src/main/java/org/duckdb/DuckDBPreparedStatement.java#L428-L430

Cc @guizmaii

staticlibs:
@jivanic-demystdata 

This should be fixed by the combination of #245 and #247.

PR #247, opened 2025-06-03
Add support for query timeouts

staticlibs:
This change implements `Statement#setQueryTimeout()` method. It is implemented by scheduling a background task and calling `Statement#cancel()` when timeout expires.

Timeouted statement has the same behaviour as it would be if cancelled manually - `SQLException` is thrown and the statement is closed.

Timeout is applied for all `execute*` calls. For `executeBatch()` it is applied separately for every single query in a batch.

Testing: new test added.

Fixes: #212

### DIFF
--- a/src/main/java/org/duckdb/DuckDBDriver.java
+++ b/src/main/java/org/duckdb/DuckDBDriver.java
@@ -5,6 +5,8 @@ import static org.duckdb.JdbcUtils.*;
 import java.nio.ByteBuffer;
 import java.sql.*;
 import java.util.*;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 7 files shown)


src/main/java/org/duckdb/DuckDBDriver.java
```diff
@@ -20,6 +21,8 @@ public class DuckDBDriver implements java.sql.Driver {
     static final String DUCKDB_URL_PREFIX = "jdbc:duckdb:";
     static final String MEMORY_DB = ":memory:";
 
+    static final ScheduledThreadPoolExecutor scheduler = new ScheduledThreadPoolExecutor(1);
```
arouel, line 24: It would be helpful when the thread name has a meaningful name. We could achieve this by using a `ThreadFactory` as second argument to set the name a new thread is created.

staticlibs: Thanks for the review! Good idea, updated the PR to add this.

Issue #84, opened 2024-09-16
[Appender] Add AppendDefault


robotslacker:
https://github.com/duckdb/duckdb/pull/11905  has merged in 1.1.0.
But in JDBC, this method is still not available.

So, can you add it?

Issue #249, opened 2025-06-03
Error when appending to DuckLake table


arouel:
When I want to append many records to a DuckLake table with an appender, I get the following error when closing the `DuckDBAppender` instance. I used `duckdb-java` version 1.3.0.0.
```
java.sql.SQLException: INTERNAL Error: Calling GetStorage on a TableCatalogEntry that is not a DuckTableEntry
This error signals an assertion failure within DuckDB. This usually occurs due to unexpected conditions or errors in the program's logic.
For more information, see https://duckdb.org/docs/stable/dev/internal_errors
	at org.duckdb.DuckDBNative.duckdb_jdbc_appender_close(Native Method)
	at org.duckdb.DuckDBAppender.close(DuckDBAppender.java:102)
	at test.DuckLakeAppenderTest.fails_appender_ducklake(DuckLakeAppenderTest.java:32)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
```

The following test cases may help to reproduce the error.
```java
package test;

import static org.assertj.core.api.BDDAssertions.*;

import java.sql.DriverManager;
import java.sql.SQLException;

import org.duckdb.DuckDBConnection;
import org.junit.jupiter.api.Test;

class DuckLakeAppenderTest {

    @Test
    void fails_appender_ducklake() throws SQLException {
        try (
            var connection = (DuckDBConnection) DriverManager.getConnection("jdbc:duckdb:");
            var statement = connection.createStatement()
        ) {
            // not necessary with duckdb 1.3.1
            connection.setAutoCommit(false);

            statement.execute("""
                ATTACH 'ducklake:/tmp/test_fails_appender_ducklake.ducklake' AS my_ducklake;
                USE my_ducklake;
                CREATE TABLE IF NOT EXISTS test (a INT);
                """);

            try (var appender = connection.createAppender("main", "test")) {
                appender.beginRow();
                appender.append(124);
                appender.endRow();
                // fails: INTERNAL Error: Calling GetStorage on a TableCatalogEntry that is not a DuckTableEntry
            }

            try (var rs = statement.executeQuery("select * from test")) {
                rs.next();
                then(rs.getInt(1)).isEqualTo(124);
            }
        }
    }

    // workaround via temporary table
    @Test
[...]

Issue #163, opened 2025-03-15
Proper way to insert arrays


blacelle:
I'm being provided a bunch of `String[]` and `double[]` and I'm looking for the proper way to insert them efficiently into a DuckDB table.

I succeed inserting (primitive) arrays only by generating equivalent concatenated String. Am I missing a way to do so?

Relates with [aggregating arrays](https://github.com/duckdb/duckdb/discussions/16670)

---

Appender: https://github.com/duckdb/duckdb-java/issues/110 confirms this is not supported by the Appender yet. The only way seems to going through a concatenated String:

```
DuckDBConnection duckDbConnection = (DuckDBConnection) c;

Statement s = duckDbConnection.createStatement();

s.execute("CREATE TABLE someTableName (color VARCHAR, array_keys STRING[%s], array_values DOUBLE[%s]);"
		.formatted(arrayLength, arrayLength));

try (var appender = duckDbConnection.createAppender(DuckDBConnection.DEFAULT_SCHEMA, tableName)) {
	for (int rowIndex = 0; rowIndex < 16; rowIndex++) {

		double rowFactor = Math.sqrt(rowIndex);

		appender.beginRow();
		appender.append("red");
		appender.append(IntStream.range(0, arrayLength).mapToObj(i -> "d" + i).collect(Collectors.joining(", ", "[", "]")));
		appender.append(IntStream.range(0, arrayLength).mapToDouble(i -> rowFactor * Math.sqrt(i)).mapToObj(d -> Double.toString(d)).collect(Collectors.joining(", ", "[", "]")));
		appender.endRow();
	}
}
```

---

BatchWriter: https://duckdb.org/docs/stable/clients/java.html#batch-writer I can not get it working:

```
DuckDBConnection duckDbConnection = (DuckDBConnection) c;

Statement s = duckDbConnection.createStatement();

s.execute("CREATE TABLE someTableName (color VARCHAR, array_keys STRING[%s], array_values DOUBLE[%s]);"
		.formatted(arrayLength, arrayLength));

PreparedStatement stmt = duckDbConnection
		.prepareStatement("INSERT INTO someTableName (color, array_keys, array_values) VALUES (?, ?, ?);");

stmt.setObject(1, "red");
stmt.setObject(2, IntStream.range(0, 255).mapToObj(i -> "d" + i).toArray(String[]::new));
stmt.setObject(3, IntStream.range(0, 255).mapToDouble(i -> Math.sqrt(i)).toArray());
[...]

Issue #157, opened 2025-03-08
need to support jdbc appender  append_default_value


Colory:
according to the official website documents and Java code.

the way of inserting the default value that is missing for the appender.

PR #295, opened 2025-06-29
Appender rework

staticlibs:
This change implements the access to Appender interface from Java with the following features:

 - C API is used to access the native Appender
 - necessary C API calls are exposed to Java using JNI wrappers as thin as possible - Java calls mirror corresponding C API calls 1 to 1
 - the data chunk interface of the Appender API is used: vector data is exposed as a direct ByteBuffer, all primitive appended values are written to this buffer from Java without going through JNI + C API (that is still used for some types with calls like: `duckdb_vector_assign_string_element_len`)
 - Java-side Appender/DataChunk/Vector data structures follow the Go lang's Appender implementation closely (with nested arrays initialization etc)
 - Java Appender usage is made thread-safe for concurrent Appender or Connection closure; `append()` calls are remained not thread-safe (to minimize the overhead), but their usage cannot cause the JNI-side crash
 - Java API of the new Appender is modeled after the `java.lang.StringBuilder` class and intended to be used with method chaining
 - support primitive arrays (one and two dimensional)
 - support for `NULL` and `DEFAULT` values
 - type checks between Java types and DB column types are enforced

Note: previous version of the Appender (that internally creates `Value` instances and appends them one by on) is still available for compatibility. It can be created using `Connection#createSingleValueAppender` method. It is marked as 'deprecated' and intended to be removed in future versions.

Testing: existing Appender test suite is extended and adapted to new Appender API.

Fixes: #84, #100, #110, #139, #157, #163, #219, #249


staticlibs:
@arouel 

The main target for the JDBC driver is Java 8. Access to DuckDB engine through FFM can be introduced in future as an optional alternative, but JNI remains the main approach for any foreseeable future. And upcoming restrictions to native code are likely to be applied (by the JDK) to both JNI and FFM the same way (because they have almost the same problems with app "integrity"). And both JNI and FFM are likely to use the same flags/manifests to disable the restrictions that are added.

arouel (commented):
Is it the plan to keep the current method names of the new `DuckDBAppender` as they were before? If possible, I would prefer if we change the naming pattern, so that one can understand which database type we append to, e.g. `appendBlob(byte[] values)`, `appendString(byte[] value)`, `appendTimestamp(Instant value)` etc.

When we keep the naming pattern as is, we keep the confusion to which type we map to.

arouel:
> @arouel
> 
> The main target for the JDBC driver is Java 8. Access to DuckDB engine through FFM can be introduced in future as an optional alternative, but JNI remains the main approach for any foreseeable future. And upcoming restrictions to native code are likely to be applied (by the JDK) to both JNI and FFM the same way (because they have almost the same problems with app "integrity"). And both JNI and FFM are likely to use the same flags/manifests to disable the restrictions that are added.

Ok, I get it, when you target Java 8 specifically.

staticlibs:
@arouel 

> Is it the plan to keep the current method names

New Appender follows the `StringBuilder`'s approach with overloaded `append()` method. When the same Java type can be used for different DB type - then special names are used like `appendDecimal()`.

Issue #298, opened 2025-07-06
[DuckDB-1.3.1.0] Geometry Functions return now `DuckDBBlobResult` instead of `Text`


manticore-projects:
```sql
select ST_GeomFromGeoJSON('{
      "type": "Point",
      "coordinates": [30.0, 10.0]}') as p;
```

with DuckDB-1.2: 
```
POINT (30 10)
```

with DuckDB-1.3.1.0:
```
DuckDBBlobResult{buffer=java.nio.HeapByteBuffer[pos=0 lim=32 cap=32]}
```

Of course I can cast to text explicitly but I believe the JDBC driver should map `GEOMETRY` to `TEXT` instead of `BLOB`.

PR #300, opened 2025-07-06
Deserialize geometry

manticore-projects:
Greetings!

This fixes the `ResultSet.getString()` method for `GEOMETRY` and issue #298:

```sql
select ST_GeomFromGeoJSON('{
      "type": "Point",
      "coordinates": [30.0, 10.0]}') as p;
```

since 1.2.2 was shown as:
```
DuckDBBlobResult{buffer=java.nio.HeapByteBuffer[pos=0 lim=32 cap=32]}
```

now with the deserialization applied it becomes again
```
POINT (30 10)
```

(Beside that I fixed two `TIMEZONE` related tests that would not work in other timezones (e.g. failed when system timezone is `ASIA/BANGKOK`).

### DIFF
--- a/src/main/java/org/duckdb/DuckDBGeometryDeserializer.java
+++ b/src/main/java/org/duckdb/DuckDBGeometryDeserializer.java
@@ -1,0 +1,691 @@
+package org.duckdb;
+
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.sql.Blob;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.stream.Stream;
+
+/**
+ * Standalone Java class for deserializing GEOMETRY BLOBs into Well-Known Text (WKT) format.
+ * Supports common geometry types: POINT, LINESTRING, POLYGON, MULTIPOINT, MULTILINESTRING,
+ * MULTIPOLYGON.
+ *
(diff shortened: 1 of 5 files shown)


staticlibs:
Hi, thanks for the PR! You can fix formatting problems by running: 

```
python ./scripts/format.py
```

Though I am not sure that adding `DuckDBGeometryDeserializer` to the driver itself is a good idea, in general we try not to have any extension-specific code in JDBC. I wonder if this logic can go into client app/connector part instead?

manticore-projects:
Greetings!

Thank you for your feedback. I have applied the formatter.

On the implementation:

- I don't think it belongs to the client because client because client will normally see `Object` or `Blob` or `String` and then simply does not know what to do with it. Also, until 1.2.2 everything was working fine. What broke it is the `ResultSet.getString()` applying `Object.toString()` on just everything including the `Blob`.
- DuckDB C code has a similar de-serializer, its just not wrapped/ported (and so I provided the port)
- I am not sure, what exactly you mean by `connector` part

I agree with you that the need for the de-serializer is sub-optimal. But only we support/implement the `GEOMETRY` type in java (which would make things worse) I really don't know how else to do it.

staticlibs:
@manticore-projects

> I am not sure, what exactly you mean by connector part

I meant some kind of intermediate code like `org.jkiss.dbeaver.ext.duckdb.model.DuckDBSQLDialect`.

I will follow-up on this after `1.3.2` release next week.

manticore-projects:
Warm reminder please, when can we discuss this? Its a big deal breaker for working with DuckDB/Spatial and SQL Tools.

PR #324, opened 2025-07-28
Add UUID support to `DuckDBAppender` (fixes #313)

### LINKED ISSUE #313: Error: unsupported C API type: 27 (UUID)
When working with the DuckLake extension, and appending data to a table with a UUID column I get the error: `unsupported C API type: 27`.

Ideally, the error message would tell a user the type name. Besides that, it would be great to support appending to UUID columns as well.

```java
package test;

import static org.assertj.core.api.BDDAssertions.*;

import java.sql.DriverManager;
import java.sql.SQLException;
import java.util.UUID;

import org.duckdb.DuckDBConnection;
import org.junit.jupiter.api.Test;

class DuckLakeAppenderTest {

    @Test
    void uuid_fails_with_new_appender_ducklake() throws SQLException {
        try (
            var connection = (DuckDBConnection) DriverManager.getConnection("jdbc:duckdb:");
            var statement = connection.createStatement()
        ) {
            statement.execute("""
                ATTACH 'ducklake:/tmp/test_uuid.ducklake' AS my_ducklake;
                USE my_ducklake;
                CREATE TABLE IF NOT EXISTS test (id UUID);
                """);

            var uuid = UUID.randomUUID().toString();
            try (var appender = connection.createAppender("main", "test")) {
                appender.beginRow();
                appender.append(uuid);
                appender.endRow();
                // java.sql.SQLException: Appender error, catalog: 'null', schema: 'main', table: 'test', message: unsupported C API type: 27
            }

            try (var rs = statement.executeQuery("select * from test")) {
                rs.next();
                then(rs.getString(1)).isEqualTo(uuid);
            }
        }
    }
}
```

arouel:
Previously, attempting to append a UUID would result in an `unsupported C API type: 27` error which is documented in https://github.com/duckdb/duckdb-java/issues/313. This change enables the `DuckDBAppender` to handle `java.util.UUID` objects directly.

### DIFF
--- a/src/main/java/org/duckdb/DuckDBAppender.java
+++ b/src/main/java/org/duckdb/DuckDBAppender.java
@@ -52,6 +52,7 @@ public class DuckDBAppender implements AutoCloseable {
 
         supportedTypes.add(DUCKDB_TYPE_ARRAY.typeId);
         supportedTypes.add(DUCKDB_TYPE_STRUCT.typeId);
+        supportedTypes.add(DUCKDB_TYPE_UUID.typeId);
     }
     private static final CAPIType[] int8Types = new CAPIType[] {DUCKDB_TYPE_TINYINT, DUCKDB_TYPE_UTINYINT};
     private static final CAPIType[] int16Types = new CAPIType[] {DUCKDB_TYPE_SMALLINT, DUCKDB_TYPE_USMALLINT};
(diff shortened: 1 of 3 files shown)


MikeSchulze (approved):

staticlibs:
@arouel 

> for what we need a Lock, because a DuckDBAppender cannot be used by multiple threads anyways

The Appender is not intended to append the data concurrently, but it can be closed concurrently. For example, if the Appenders's connection is closed, the Appender is closed automatically. This connection close can well be triggered concurrently, for example from the shutdown cleanup code. After the Appender is closed - its native part is destroyed. And the destruction of the native part cannot be concurrent with it usage from other thread. Almost all `append()` methods (with the notable exception of strings over 12 bytes long) are Java-local - they only lay the input data out into the Data Chunk (thus they should be fast). But `flush()` operation calls to the native part and cannot be concurrent with `close()`. So the `appenderRefLock` is used to put them in order.

arouel:
Thank you for the thorough response.

PR #353, opened 2025-08-27
Backport #18374 engine fix to v1.3.2 branch

staticlibs:
This PR contains a cherry-pick of the duckdb/duckdb#18752 engine PR applied on top of the `v1.3.2.0` release. `v1.3.2` branch is created for it so the new release built from this branch will be compatible with existing `v1.3.2` extensions.

Testing: test case included with the original PR is run manually on patched JDBC.


staticlibs:
@lnkuiper I wonder if you can take a look at this backport (that it contains all intended bits)? I've checked manually that the query [from this comment](https://github.com/duckdb/duckdb/pull/18374#pullrequestreview-3087071254) does not produce any results on patched JDBC.

Issue #180, opened 2025-03-29
Graal Native Image Support


aherlihy:
Hi, does the DuckDB JDBC driver support use from a graal native image?

I tried with v1.1.1 using graalvm-jdk-17.0.12+8.1 using SBT NativeImagePlugin, but it fails with:

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: Can't load library: <...>/build/debug/libduckdb_java.so_osx_universal
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.NativeLibrarySupport.loadLibraryAbsolute(NativeLibrarySupport.java:100)
	at java.base@17.0.12/java.lang.ClassLoader.loadLibrary(ClassLoader.java:57)
	at java.base@17.0.12/java.lang.Runtime.load0(Runtime.java:755)
	at java.base@17.0.12/java.lang.System.load(System.java:1957)
	at org.duckdb.DuckDBNative.<clinit>(DuckDBNative.java:52)
	at org.duckdb.DuckDBConnection.newConnection(DuckDBConnection.java:51)
	at org.duckdb.DuckDBDriver.connect(DuckDBDriver.java:41)
	at java.sql@17.0.12/java.sql.DriverManager.getConnection(DriverManager.java:681)
	at java.sql@17.0.12/java.sql.DriverManager.getConnection(DriverManager.java:252)
        ...
```
on Ubuntu 22.04 LTS (and equivalently `build/debug/libduckdb_java.so_osx_universal` on MacOS 15.3.1). It works fine from a regular JAR. The build.sbt native image settings are:

```
 .settings(
    nativeImageInstalled := true,
    nativeImageGraalHome := {
      val executable = "native-image"
      val paths = System.getenv("PATH").split(Pattern.quote(pathSeparator))
      paths.map(Paths.get(_)).find(p => Files.isExecutable(p.resolve(executable))) match {
        case Some(binDir) =>
          binDir.getParent
        case None =>
          throw new MessageOnlyException(s"Could not find binary `$executable` in PATH.")
      }
    },
    nativeImageCommand := nativeImageCommand.value ++ Seq(
      "--no-fallback",
      "--report-unsupported-elements-at-runtime"
    ),

    libraryDependencies ++= Seq(
       ...,
      "org.duckdb" % "duckdb_jdbc" % "1.1.1",
    )
```

[...]

PR #421, opened 2025-10-12
Load DuckDB native library by name first

hubertp:
When loading a native library, attempt to first load the library, `duckdb_java`, by name. Only when that fails, loading will revert to the old behaviour of searching via classpath/local file path.

The main motivation for this work was to allow creating a correct [Native Image](https://www.graalvm.org/latest/reference-manual/native-image/) that will not have issues during initialization. The change does not provide the support out-of-the-box.
But it works in Native Image if someone (like [us](https://github.com/enso-org/enso/pull/14114)) extracts the native library (wrt OS/Arch), renames it to follow the specification and places it on JNI search path.

For `System.loadLibrary` to work, the native library has to have the correct prefix (`lib` for Mac/Linux, no `lib` for Windows) and suffix (`.so`, `.dll` or `.dylib` for Linux/Windows/Mac, respectively). No architecture suffix is allowed.
As it stands the current packaging logic does not create native library files with the right names but that seems like a deeper change in the build setup. Also, I have zero knowledge about how the artifacts are being used internally, in dev work, so I don't feel comfortable with changing it without guidance.
One could look at [sqlite-jdbc](https://mvnrepository.com/artifact/org.xerial/sqlite-jdbc) for some inspiration on how to improve it (which is roughly what we had to do when repackaging DuckDB's Java client jar).

Also, if maintainers are fine with this change we can also include [Native Image configuration](https://github.com/enso-org/enso/pull/14114/files#diff-4116f4f91b7bd32d3ab9ef6c79e25f1ee34782039dbcfe4cab8760d099751a00) in the final `duckdb_jdbc.jar` which is a requirement for GraalVM's Native Image. More and more Java libraries provide similar Native Image configuration.
That would mostly enable https://github.com/duckdb/duckdb-java/issues/180.


hubertp:
> @hubertp
> 
> Thanks for updating the PR! Just FYI, this change is intended to be included into 1.4.2 update planned for Nov 10.

Thanks, I will create a separate PR with Native Image properties.

staticlibs:
@hubertp 

I took the idea from this PR a little bit further and filed #450, I wonder if you can take a look at it whether it (along with #447) covers your requirements?

src/main/java/org/duckdb/DuckDBNative.java
```diff
@@ -15,38 +15,47 @@
...
-            } else if (os_name_detect.startsWith("mac")) {
-                os_name = "osx";
-                os_arch = "universal";
-            } else if (os_name_detect.startsWith("linux")) {
-                os_name = "linux";
-            }
-            String lib_res_name = "/libduckdb_java.so"
-                                  + "_" + os_name + "_" + os_arch;
+        String libname = "duckdb_java";
+        boolean loaded = loadLibraryByName(libname);
```
JaroslavTulach: - as far as I can say, the PR:
   - asks the JVM to load the library by name first
   - only if that fails, it "tries harder" to find the library itself
- using `System.loadLibrary` is [runtime container friendly](https://wiki.apidesign.org/wiki/NetBeans_Runtime_Container) and packagers of full applications (like [Enso](http://github.com/enso-org/enso)) friendly
- "trying harder" is _"Maven friendly"_ - enough to add dependency to `pom.xml` and things work

staticlibs:
@hubertp 

Just FYI, I've filed an experimental PR with `reachability-metadata.json` - #452 (where the JSON file is taken from the link in OP verbatim).

I wonder if there are any meaningful ways to test this metadata with the `native-image`? As the `duckdb_java_test.jar` test suite is not runnable with it, I wonder if some kind of a "smoke check" can be run with a `native-image` and the main JAR?

PR #515, opened 2026-01-08
Add TIME_NS support

staticlibs:
This PR adds support for fetching `TIME_NS` values as instances of `java.time.LocalTime` class, example:

```java
LocalTime lt = rs.getObject(1, LocalTime.class);
```

Nanosecond precision is maintained on fetching `TIME_NS` values from DB.

Currently it is not possible to preserve nanosecond precision when inserting `TIME_NS` values using query parameters.

The support is necessary to fix `test_all_types` failures introduced by duckdb/duckdb#20361.


staticlibs:
Merging to allow vendoring PR #513 to pass tests and be integrated, bringing in updated engine with duckdb/duckdb#20361 included.

PR #551, opened 2026-01-30
Fix optimizer crash in JDBC by ensuring active_query is set during prepare

thijs-s:
This PR fixes a crash/NPE encountered when using optimizer extensions with the JDBC driver.

### Issue
When `Connection.prepareStatement()` is called, `ClientContext::PrepareInternal` is executed. Previously, `active_query` was not set in the `ClientContext` during this phase. If an optimizer extension (hooked into `PreparedStatement` creation) attempted to access `context.GetCurrentQuery()`, it would dereference a null pointer (or hit an assertion in debug builds), causing a crash.

### Solution
The `ClientContext` requires an initialized `active_query` state for the optimizer to function correctly (e.g., to access the query string). This state was previously missing during the JDBC prepare phase (`PrepareInternal`), which caused the crash.

We now correctly initialize `active_query` within `PrepareInternal` using a new RAII helper struct `ActiveQueryGuard`. This ensures the context is fully populated for the duration of the preparation, satisfying the requirements of optimizer extensions.

- The `ActiveQueryGuard` sets `active_query` upon initialization.
- It guarantees `active_query` is reset to `nullptr` when the scope exits, preserving `ClientContext` invariants.
- The implementation is exception-safe: even if an exception occurs during statement preparation or memory allocation, the destructor ensures proper cleanup.


thijs-s:
thanks for the info! Done. https://github.com/duckdb/duckdb/pull/20755

PR #566, opened 2026-02-12
Fix SNAPSHOT metadata: deploy POM and JAR together

guizmaii:
## Summary

- Fix SNAPSHOT dependency resolution failing in Coursier/sbt due to mismatched build numbers between POM and JAR artifacts

## Problem

Each `mvn deploy:deploy-file` call creates a new build number in the SNAPSHOT `maven-metadata.xml`. The current script deploys POM and JAR as separate calls:

```
deploy POM     → build 1 (timestamp 231930)
deploy JAR     → build 2 (timestamp 231938)
deploy sources → build 3 (timestamp 231947)
...
```

Dependency resolvers like Coursier (used by sbt/Scala) pick the JAR's build number from the metadata and then try to find the POM at the **same** build number. Since the POM was deployed with build number 1 but Coursier looks for it at build number 2, the download returns 404:

```
# Coursier tries this (JAR's build number):
GET .../duckdb_jdbc-1.5-627976eb-20260211.231938-2.pom  → 404

# But POM is actually here (different build number):
GET .../duckdb_jdbc-1.5-627976eb-20260211.231930-1.pom  → 200
```

## Fix

Deploy the main JAR with `-DpomFile=<path>` instead of deploying the POM separately. This way Maven publishes both artifacts in a single operation, giving them the same build number and timestamp.

## Test plan

- [ ] Verify SNAPSHOT artifacts are published correctly after merge
- [ ] Verify `sbt` / Coursier can resolve the SNAPSHOT dependency without errors


staticlibs:
Thanks! It looks good to me.

PR #601, opened 2026-03-16
Add big endian support to v1.5

DNikolaevAtRocket:
The PR adds big-endian support to the JDBC driver v1.5.

Big-endian patches landed to DuckDB v1.5 (see https://github.com/duckdb/duckdb/pull/19748 and https://github.com/duckdb/duckdb/pull/20237); but the driver needs several minor updates.


staticlibs:
Thanks!

Issue #605, opened 2026-03-17
Running driver on not officially supported platforms


DNikolaevAtRocket:
### Description

It's not possible to run the driver on platforms that are not officially supported.

- Regardless of which version of `duckdb_jdbc_*` is used, it calls `cpuArch` and `osName` unconditionally before it tries to load plain `duckdb_java`.
Each of them throws `IllegalStateException` if being run not on one of the hardcoded OS's / architectures.
- There's `OVERRIDE_JDBC_OS_ARCH` to enforce the correct arch for the native part, but there's no such option for `OS_NAME`.

I'd be happy to provide a PR for this.
Any ideas on how to address it?
From my side, I can see the below options.

#### For the native part

Introduce `OVERRIDE_JDBC_OS_NAME`.

#### Option 1 for the Java part (more generic)

Package a properties file that'd contain all the mappings between `os.name` / `os.arch` <-> OS / arch used by `DuckDBNative`, and load it in runtime.
It could contain all the officially supported platforms + extra platforms defined in build time / edited in pre-built jars.

E.g.
```properties
<...>
duckdb.os.arch.arm64 = aarch64,arm64
duckdb.os.name.linux = linux
<...>
```
It'd allow using the official build in the most flexible way.

#### Option 2 for the Java part (more conservative)

Catch `IllegalStateException("Unsupported <...>")`, try to load plain `duckdb_java`. If the latter fails, throw the original error.
Or something like this.
  
It'd allow using `duckdb_jdbc-*-nolib.jar` on such platforms.

 
Thank you.

staticlibs:
@DNikolaevAtRocket 

Thanks for raising the issue!

I think the option 2 is better. Just I suggest to restructure the name generation part a little to go to `System.loadLibrary()` on unsupported arches without throwing and catching the exception.

DNikolaevAtRocket:
@staticlibs,

Thanks for quick response!
Will submit a PR accordingly.

I hope it's okay if it goes to the v1.5 branch with forward porting to `main`.
Please let me know if it should go to `main` only.

staticlibs:
@DNikolaevAtRocket 

> I hope it's okay if it goes to the v1.5 branch with forward porting to main.
> Please let me know if it should go to main only.

It is fine to submit it to either 1.5 or `main`. It should go into both of them, so I will cherry-pick it to the second branch. Right now 1.5 overall contains more recent changes than `main`, so it may be convenient to work on it.

PR #606, opened 2026-03-18
Improve cross platform support

DNikolaevAtRocket:
The PR addresses https://github.com/duckdb/duckdb-java/issues/605 as it was discussed by introducing `OVERRIDE_JDBC_OS_NAME` parameter and removing the exceptions from `cpuArch` and `osName`.

On top of that, it includes a couple of minor updates for z/OS (reflecting the ones DuckDB core has).


staticlibs:
Thanks!

Issue #622, opened 2026-03-26
DuckDB 1.5.1 (JDBC via DBeaver) sets temp_directory=\\.tmp on Windows causing “IO Error: Failed to create directory '.tmp'”


aromlibero:
**What happens?**
When connecting to DuckDB 1.5.1 through JDBC using DBeaver on Windows, DuckDB assigns the temporary directory:
_temp_directory = \.tmp_

This path is invalid on Windows, resulting in:
_IO Error: Failed to create directory "\.tmp": Invalid path_

The same setup with DuckDB 1.4.4 correctly sets:
_temp_directory = \mydatabase.duckdb.tmp_

so no error occurs.

**Expected behavior**
DuckDB should assign a valid writable temporary directory on Windows, similar to:
_temp_directory = <db_file>.tmp_

as done in version 1.4.4, or fall back to a correct system temp folder (e.g. %TEMP%).

**Observed behavior**
In version 1.5.1:
- temp_directory becomes \\.tmp
- Windows interprets \\.tmp as an invalid UNC path
- Any operation requiring temp spill fails with an IO error
This behavior matches known issues where DuckDB attempts to create a .tmp directory in read‑only or invalid locations.

**Steps to reproduce**
Windows 11
DBeaver
- Create a connection using DuckDB JDBC driver 1.5.1
- Open SQL editor and run:
_SELECT * FROM duckdb_settings() WHERE name='temp_directory';_

Observe the value \\.tmp
Run any query requiring temporary storage (e.g. large intermediate sort)
DuckDB fails with:
_IO Error: Failed to create directory "\.tmp"_

**DuckDB versions tested**
1.4.4 → works correctly
1.5.1 → temp_directory=\.tmp (invalid)

**Environment**
OS: Windows 11
Client: DBeaver + JDBC
Database: Single-file DuckDB
Filesystem: Local, fully writable

**Workaround**
Manually setting:
_SET temp_directory='C:/Users/<user>/AppData/Local/Temp';_
fixes the issue, confirming the problem is in DuckDB's automatic temp directory resolution.

**Request**
Please restore the previous behavior or implement a Windows‑safe fallback when computing temp_directory, so JDBC users on Windows do not receive invalid paths like \\.tmp.

staticlibs:
Hi, thanks for the report! Seems to be an unintended behaviour on Windows, I will take a look.

staticlibs:
The problem appeared to be Windows-specific and Java-specific (CLI is unaffected).

[...]

PR #625, opened 2026-03-27
Do not provide custom FS in config (1.5)

staticlibs:
This is a backport of the PR #624 to `v1.5-variegata` stable branch.

Workaround with providing custom FS instance as part of initial DB config is no longer necessary in recent versions. In the past it was necessary to use `allowed_directories` option. This PR removes the custom FS.

Testing: the change is covered by existing test.

Ref: #622


staticlibs:
Appeared to cause other config problems on DB startup, closing as not-planned.

PR #658, opened 2026-04-13
Add Jepsen to CI

staticlibs:
This PR adds the [DuckDB Jepsen test suite](https://github.com/jepsen-io/duckdb) to test runs on CI.


staticlibs:
Just if someone is curious, Jepsen test suite currently fails on main because duckdb/duckdb#21489 fix has not yet landed on main. It runs fine on 1.5 in #660.

PR #663, opened 2026-04-15
ISSUE-22042: JDBC: fix PIVOT with auto-commit off (multi-statement prepare)

### LINKED ISSUE #22042: Regression: DuckDB JDBC 1.5.2.0 can't execute PIVOT statements anymore with AUTO COMMIT false
Given

```java
import java.sql.DriverManager;
import java.sql.SQLException;

public class F {
	public static void main(String... a) throws SQLException {
		try (var connection = DriverManager.getConnection("jdbc:duckdb:");) {
			connection.setAutoCommit(false);
			try (var stmt = connection.createStatement()) {

				var result = stmt.executeQuery("pivot values(1,2),(1,3) x(y,z) on x using first(y)");
				while(result.next()) {
					System.out.println(result.getObject(1) + ", " + result.getObject(2));
				}
				result.close();
			}
		}
	}
}
```

With DuckDB 1.5.1.0, results is:

```
2, 1
3, null
```

DuckDB 1.5.2.0 fails immediately with

```
Exception in thread "main" java.sql.SQLException: executeQuery() can only be used with queries that return a ResultSet
	at org.duckdb.DuckDBPreparedStatement.executeQuery(DuckDBPreparedStatement.java:278)
	at org.duckdb.DuckDBPreparedStatement.executeQuery(DuckDBPreparedStatement.java:310)
	at F.main(F.java:10)
```

herbenderbler:
Fixes https://github.com/duckdb/duckdb/issues/22042

## Problem

With **JDBC 1.5.2+** and **`Connection.setAutoCommit(false)`**, executing a **dynamic PIVOT** (pivot values inferred from data, so DuckDB expands the SQL into several statements) via **`Statement.executeQuery`** failed immediately with:

`java.sql.SQLException: executeQuery() can only be used with queries that return a ResultSet`

The engine’s statement preprocessor can wrap multi-statement bodies in an active transaction with leading/trailing **`SET current_transaction_invalidation_policy`** statements. The JNI path executed all statements except the last with `Query`, then **`Prepare`’d only `statements.back()`**. When the syntactic last statement was that trailing **`SET`**, the prepared “query” did not return a **`QUERY_RESULT`**, so JDBC correctly rejected **`executeQuery`**. Starting a JDBC transaction **before** `duckdb_jdbc_prepare` also made that expansion more likely at prepare time.

## Solution

- In **`duckdb_jdbc_prepare`** and **`duckdb_jdbc_pending_query`**, locate the **last `SELECT_STATEMENT`** in the extracted statement list, run all **prefix** statements with `Query`, **`Prepare` / `PendingQuery`** that SELECT, and store **trailing** SQL text to run **after** each successful **`Execute`** (so repeated `execute()` on the same `PreparedStatement` still runs policy resets).
- In **`DuckDBPreparedStatement.prepare()`**, **do not** call **`startTransaction()`** before native prepare; transactions are still started in **`execute()`** when **`autoCommit`** is off.

## Changes

- `src/jni/duckdb_java.cpp` — `FindLastSelectStatementIndex`, `ExecuteTrailingQueries`; updated prepare / pending / execute paths.
- `src/jni/holders.hpp` — `StatementHolder::trailing_queries_after_execute`; `PendingHolder` trailing queries + `connection_for_trailing`.
- `src/main/java/org/duckdb/DuckDBPreparedStatement.java` — remove `startTransaction()` from `prepare()`.
- `src/test/java/org/duckdb/TestDuckDBJDBC.java` — `test_issue_22042_pivot_autocommit_false`.

## Testing

- `python3 scripts/format.py` (project formatter).
- Build: `make release` (or your usual CMake build), producing `build/release/duckdb_jdbc.jar` and `build/release/duckdb_jdbc_tests.jar`.
- Automated: `java -cp "build/release/duckdb_jdbc_tests.jar:build/release/duckdb_jdbc.jar" org.duckdb.TestDuckDBJDBC test_issue_22042_pivot_autocommit_false`
- Related sanity: `test_autocommit_off`, `test_multiple_statements_execution` via the same `TestDuckDBJDBC` runner.

#### 1) Reproduce the original error (pre-fix behavior)

Use a **JDBC driver build from before this patch** (e.g. **1.5.2.0** as in the report, or a local build of `main` without these commits). Ensure **`duckdb_jdbc.jar`** and **`libduckdb_java`** (or the platform JNI library) are on the classpath / library path as usual for your environment.

1. Save as `F.java`:

```java
import java.sql.DriverManager;
import java.sql.SQLException;

public class F {
	public static void main(String... a) throws SQLException {
		try (var connection = DriverManager.getConnection("jdbc:duckdb:")) {
			connection.setAutoCommit(false);
			try (var stmt = connection.createStatement()) {
				var result = stmt.executeQuery(
					"pivot values(1,2),(1,3) x(y,z) on x using first(y)");
				while (result.next()) {
					System.out.println(result.getObject(1) + ", " + result.getObject(2));
				}
				result.close();
			}
		}
	}
}
```

2. Compile and run (adjust classpath to your pre-fix **`duckdb_jdbc.jar`**):

```bash
javac -cp duckdb_jdbc.jar F.java
java -cp ".:duckdb_jdbc.jar" F
```

**Expected (bug):** immediate failure with:

`java.sql.SQLException: executeQuery() can only be used with queries that return a ResultSet`

With **`setAutoCommit(true)`** (default), the same program typically **succeeds** and prints two rows (e.g. `2, 1` and `3, null`), which highlights the regression tied to **manual commit / active transaction** semantics.

#### 2) Confirm this patch works (post-fix)

1. Build the driver from this branch:

```bash
cd /path/to/duckdb-java
make release
```

2. Run the regression test:

```bash
java -cp "build/release/duckdb_jdbc_tests.jar:build/release/duckdb_jdbc.jar" \
  org.duckdb.TestDuckDBJDBC test_issue_22042_pivot_autocommit_false
```

**Expected:** `TestDuckDBJDBC#test_issue_22042_pivot_autocommit_false success` and `OK`.

3. Re-run the **`F.java`** sample above, but put **`build/release/duckdb_jdbc.jar`** on the classpath (and ensure the JNI library from the same build is loaded, e.g. via `-Djava.library.path=build/release` if required by your setup).

**Expected:** no exception; two rows printed (values equivalent to **`2, 1`** and **`3, null`**, order may vary).

### DIFF
--- a/src/jni/duckdb_java.cpp
+++ b/src/jni/duckdb_java.cpp
@@ -190,7 +190,35 @@ void _duckdb_jdbc_disconnect(JNIEnv *env, jclass, jobject conn_ref_buf) {
 	}
 }
 
+#include "duckdb/common/constants.hpp"
+#include "duckdb/common/enums/statement_type.hpp"
 #include "utf8proc_wrapper.hpp"
+
+namespace {
+
+//! When a query is expanded into several statements (e.g. dynamic PIVOT creating a temp enum, or transaction policy SET
+//! wrappers), the JDBC driver must prepare the statement that actually returns the result set — typically the last
+//! SELECT — not merely the syntactically last statement.
+idx_t FindLastSelectStatementIndex(const duckdb::vector<duckdb::unique_ptr<SQLStatement>> &statements) {
+	for (idx_t i = statements.size(); i > 0; i--) {
+		if (statements[i - 1]->type == StatementType::SELECT_STATEMENT) {
+			return i - 1;
+		}
+	}
+	return DConstants::INVALID_INDEX;
+}
+
+void ExecuteTrailingQueries(ClientContext &context, const duckdb::vector<string> &trailing_queries) {
+	QueryParameters parameters;
+	for (const auto &trailing_sql : trailing_queries) {
+		auto res = context.Query(trailing_sql, parameters);
+		if (res->HasError()) {
+			res->ThrowError();
+		}
+	}
+}
+
+} // namespace
 
 jobject _duckdb_jdbc_prepare(JNIEnv *env, jclass, jobject conn_ref_buf, jbyteArray query_j) {
 	auto conn_ref = get_connection(env, conn_ref_buf);
@@ -205,9 +233,17 @@ jobject _duckdb_jdbc_prepare(JNIEnv *env, jclass, jobject conn_ref_buf, jbyteArr
 		throw InvalidInputException("No statements to execute.");
 	}
 
-	// if there are multiple statements, we directly execute the statements besides the last one
-	// we only return the result of the last statement to the user, unless one of the previous statements fails
-	for (idx_t i = 0; i + 1 < statements.size(); i++) {
+	idx_t prepare_idx = FindLastSelectStatementIndex(statements);
+	if (prepare_idx == DConstants::INVALID_INDEX) {
+		prepare_idx = statements.size() - 1;
+	}
+
+	duckdb::vector<string> trailing_queries;
+	for (idx_t i = prepare_idx + 1; i < statements.size(); i++) {
+		trailing_queries.push_back(statements[i]->ToString());
+	}
+
+	for (idx_t i = 0; i < prepare_idx; i++) {
 		auto res = conn_ref->Query(std::move(statements[i]));
 		if (res->HasError()) {
 			res->ThrowError();
@@ -215,7 +251,8 @@ jobject _duckdb_jdbc_prepare(JNIEnv *env, jclass, jobject conn_ref_buf, jbyteArr
 	}
 
 	auto stmt_ref = make_uniq<StatementHolder>();
-	stmt_ref->stmt = conn_ref->Prepare(std::move(statements.back()));
+	stmt_ref->stmt = conn_ref->Prepare(std::move(statements[prepare_idx]));
+	stmt_ref->trailing_queries_after_execute = std::move(trailing_queries);
 	if (stmt_ref->stmt->HasError()) {
 		string error_msg = string(stmt_ref->stmt->GetError());
 		stmt_ref->stmt = nullptr;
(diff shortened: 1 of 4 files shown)


staticlibs:
Hi, thanks for the PR!

> Solution
> In duckdb_jdbc_prepare and duckdb_jdbc_pending_query, locate the last SELECT_STATEMENT in the extracted statement list

`SELECT_STATEMENT` are not the only ones that can return results, should we check for other types too?

> In DuckDBPreparedStatement.prepare(), do not call startTransaction() before native prepare; transactions are still started in execute() when autoCommit is off.

I think this will make `prepare` and `execute` to run in different transactions, when prepared statement is used with auto-commit disabled and no transaction is already running. In general this is undesired as extensions can use some transaction-local state when running the queries.

Please do not post LLM-generated comments.

herbenderbler:
> `SELECT_STATEMENT` are not the only ones that can return results, should we check for other types too?

Yeah, I see your point.  If we only focus on `SELECT_STATEMENT`, then this change is sufficient.

However, if we want something more robust for all future multi-statement expansions, I would have to either widen this logic or document this as the path forward.

> I think this will make `prepare` and `execute` to run in different transactions, when prepared statement is used with auto-commit disabled and no transaction is already running. In general this is undesired as extensions can use some transaction-local state when running the queries.

I see your point how currently it's assumed one explicit transaction spans both `prepare` and `execute` work.   This is a reasonable expectation for a driver which uses `BEGIN` in this manner.

The reason I touched this bit at all is because Issue #663 is tied to having an active transaction white the statement is prepared.  This changed how the engine rewrote the batch with the extra `SET current_transaction_invalidation_policy`.  Given this synthetic last statement, this isn't the `SELECT` that returns teh `ResultSet`.

> Please do not post LLM-generated comments.

No problem.  These are all of my own broken, incomplete thoughts 😆

staticlibs:
> If we only focus on SELECT_STATEMENT, then this change is sufficient.

Agree, it will work in most cases. Just the check itself looks wrong to me, we expect that the statements returned by the engine match the input statements 1 to 1. If they don't match - we'd rather have some better protocol to detect how the input list was modified/decorated rather than trying to guess it by the statement type. This logic with running all queries and returning the result of the last one is used in a few clients - not only in JDBC.

> This changed how the engine rewrote the batch with the extra SET current_transaction_invalidation_policy.

Yes, this is unfortunate and IMO should not happen in the engine. Just removing this "BEGIN" before "prepare" does not really help with the OP problem if transaction is already open (if any "SELECT" was run on this connection before).

> No problem. These are all of my own broken, incomplete thoughts 😆

Thanks :smiley:! I think this PR is quite useful, just I will wait for the feedback from the engine people before doing anything with it.

staticlibs:
Closing this in favour of this engine PR - duckdb/duckdb#22141, please leave a comment if this needs to be re-opened.

PR #738, opened 2026-07-15
feat: add JDBC instance cache option

leonardovida:
Separate JDBC connections sometimes need separate DuckDB instances so instance-level state, such as attached catalogs, cannot change underneath sibling connections. This adds an opt-in jdbc_instance_cache=false property while keeping the current sharing behavior by default.

```java
Properties properties = new Properties();
properties.setProperty("jdbc_instance_cache", "false");

try (Connection a = DriverManager.getConnection(url, properties);
     Connection b = DriverManager.getConnection(url, properties)) {
    // a and b use independent DuckDB instances
}
```


staticlibs (approved):
Thanks for the PR! Looks good to me.

PR #749, opened 2026-07-22
Propagate duckdb commit and version to Java.yml

smvv:
### Context

This PR is part of the [alpha](https://github.com/duckdblabs/duckdb-internal/issues/8628) release versions initiative.

The added action sends the release state of the python client back to the [release event system](https://github.com/duckdblabs/duckdb-internal/issues/9742).

See also the release event system [usage](https://github.com/duckdb/duckdb-workflow-trigger/#usage).

### Changes

I had to make some CI changes so we can propagate `workflow_dispatch` values to the CI run.

The flow is now:
1. duckdb core sends `core_ready` to Release event system.
2. Release event system [sends](https://github.com/duckdb/duckdb-workflow-trigger/blob/46784168c8ee1c92104fb0eec8a0107f248d917d/endpoints.yml#L9) a dispatch event to `duckdb/duckdb-java/Vendor.yml`
3. `Vendor` dispatches `Java.yml` with the required input parameters.
4. `Vendor.yml` sends `client_ready` back to Release event system.

Without the CI job/workflow restructuring, the values for `duckdb_version` and `duckdb_commit` were lost when we want to send the `client_ready` back.

### DIFF
--- a/.github/workflows/Java.yml
+++ b/.github/workflows/Java.yml
@@ -1,6 +1,8 @@
 name: Java JDBC
 on:
   push:
+    branches-ignore:
+      - 'vendoring-*'
   pull_request:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/Java.yml
```diff
@@ -18,6 +35,7 @@ jobs:
 
   format-check:
     name: Format Check
+    if: ${{ github.event_name != 'pull_request' || !startsWith(github.head_ref, 'vendoring-') }}
```
smvv, line 38: note: we're filtering on the `head_ref` (= `vendoring-*`) instead of the `base_ref` (= `main`), so `pull_request.branch-ignores` does not work, unfortunately.

PR #790, opened 2026-08-07
Cache CAPIType values

kkristof93:
`CAPIType.capiTypeFromTypeId()` called `CAPIType.values()` on every invocation, which allocates a fresh array copy of all enum constants each time. For a hot path (type resolution on every vector/column read) this creates unnecessary heap allocations and linear scan overhead.

The fix introduces a `BY_TYPE_ID` static array built once during class initialization by iterating `values()` a single time. Lookups are now a bounds check plus a single array dereference (O(1) with no allocation, compared to the previous O(n) scan with a transient array allocation per call).

### DIFF
--- a/src/main/java/org/duckdb/DuckDBBindings.java
+++ b/src/main/java/org/duckdb/DuckDBBindings.java
@@ -4,9 +4,6 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.sql.SQLException;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.LocalTime;
 
 public class DuckDBBindings {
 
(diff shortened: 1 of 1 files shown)


staticlibs (approved):
Thanks for the PR! Looks good to me. I'll add it to 1.5 branch and will publish a snapshot shortly.

staticlibs:
Snapshot with this change was published, compatible with DuckDB 1.5.5:

```xml
<dependency>
    <groupId>org.duckdb</groupId>
    <artifactId>duckdb_jdbc</artifactId>
    <version>1.5.5.2-dev-651bc51</version>
</dependency>
```
```xml
<repository>
    <id>duckdb</id>
    <url>https://duckdb-staging.duckdb.org/duckdb/duckdb-java/maven/</url>
</repository>
```

PR #803, opened 2026-08-12
Add lifecycle cleanup for table function states

### LINKED ISSUE #800: Java table functions have no deterministic cleanup hook for state objects
### Problem

`DuckDBTableFunction` allows Java table functions to return bind, global-init, and local-init state objects, but it provides no deterministic teardown hook for those objects.

If execution stops before `apply()` returns zero, such as with `LIMIT`, early `ResultSet.close()`, cancellation, or an error, resources owned by the state
cannot be closed reliably. Releasing the JNI global reference and relying on GC is insufficient for JDBC cursors, streams, iterators, or native handles.

### Reproduction

Register a table function whose global-init state opens a tracked resource, then run:

```sql
SELECT * FROM java_lifecycle_test() LIMIT 1
```
After reading one row, close both the ResultSet and the Statement.
Observed with duckdb_jdbc 1.5.5.1 and current main:
active resources: 1
close calls: 0
Expected: the state receives deterministic cleanup when DuckDB destroys the corresponding native table-function state.

### Analysis
The C API already supports delete callbacks through duckdb_bind_set_bind_data() and duckdb_init_set_init_data().
The Java JNI binding currently uses GlobalRefHolder::destroy, which only deletes the JNI global reference. Therefore, the core already knows when cleanup must occur; the missing part is a Java lifecycle contract.

### Proposed solution
Add a small opt-in marker interface:
public interface DuckDBTableFunctionState extends AutoCloseable {
    @Override
    void close() throws Exception;
}
When a bind, global-init, or local-init state implements this interface, the JNI delete callback would invoke close() before deleting the global reference.
This approach:
- is source- and binary-compatible;
- does not change ownership of existing AutoCloseable state objects;
- uses the lifecycle already provided by the DuckDB C API;
- requires no changes to the DuckDB core.
The existing generic GlobalRefHolder::destroy should remain unchanged. A state-specific destroy callback can perform best-effort cleanup, clear any JNI exception, and always release the global reference.

lfkpoa:
## Summary

Fixes #800.

Add an opt-in lifecycle contract for Java table function state objects returned
from `bind`, global init, and local init.

States implementing the new `DuckDBTableFunctionState` interface are closed when
DuckDB destroys the corresponding native state:

```java
public interface DuckDBTableFunctionState extends AutoCloseable {
  @Override
  void close() throws Exception;
}
```
Objects implementing only AutoCloseable, regular objects, and null retain their existing behavior.

## Implementation

- Add the public DuckDBTableFunctionState marker interface.
- Add an internal Java teardown helper that catches failures as best effort.
- Register a specialized JNI delete callback for bind/global/local state.
- Attach native worker threads to the JVM when required.
- Clear unexpected pending JNI exceptions and always release the global reference.

- Keep the generic GlobalRefHolder::destroy callback unchanged for table function extra info and other references.

- Document ownership, concurrency, aliasing, and exception behavior in UDF.MD.

close() is invoked once per installed native holder. Therefore, the same Java object installed in multiple holders may receive multiple calls and should be idempotent when resources are shared.

## Tests

Added coverage for:

- normal exhaustion;
- early termination with LIMIT;
- errors from apply;
- connection-driven cleanup;
- streaming result cleanup;
- repeated JDBC close operations;
- repeated executions;
- parallel local states;
- aliased state objects;
- exceptions thrown by close;
- null and regular state objects;
- AutoCloseable objects without the marker.

### DIFF
--- a/UDF.MD
+++ b/UDF.MD
@@ -218,6 +218,38 @@ Table function callback must implement `DuckDBTableFunction` interface:
  - `localInit` (optional) - called once for every thread before the function execution, returns an object used to track thread-local state during function execution
  - `apply` - has access to bind data, global init data and local init date, writes results into output vectors, returns a number of written row, is called by the engine repeatedly until it return zero rows
 
+States that own resources can opt into deterministic cleanup by implementing
+`DuckDBTableFunctionState`:
+
+```java
+final class SourceState implements DuckDBTableFunctionState {
+    private final AutoCloseable resource;
+
+    SourceState(AutoCloseable resource) {
+        this.resource = resource;
+    }
+
+    @Override
+    public void close() throws Exception {
+        resource.close();
+    }
+}
+```
+
+After a state is successfully installed as bind, global-init, or local-init
+data, DuckDB owns its lifecycle and calls `close()` when that state is
+destroyed. This can happen after normal exhaustion, a `LIMIT`, an error,
+cancellation, or early JDBC result/statement/connection closure. Only objects
+implementing `DuckDBTableFunctionState` opt in; an object implementing only
+`AutoCloseable`, and `null`, are left unchanged.
+
+Cleanup can run on native worker threads. Local states can be closed
+concurrently, and no global ordering is guaranteed between bind, global-init,
+and local states. Make `close()` fast, idempotent, and thread-safe when it
+manages shared resources. It must not execute SQL or re-enter the same
+connection. Exceptions from `close()` are handled on a best-effort basis and
+are not propagated to the query, so record any diagnostics inside the state.
+
 To enable parallel function execution it is necessary to set desired number of threads using `info.setMaxThreads()` method during the global init call.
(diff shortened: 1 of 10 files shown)


staticlibs (approved):
Thanks for the PR and for the thorough description and tests! It looks good to me.

Perhaps we can just use `AutoClosable` instead of the marker interface, but this can be extended later. And the marker interface seems to be completely non-breaking for existing code, so can be safely backported to 1.5.

lfkpoa:
I used a dedicated marker interface instead of closing every AutoCloseable state to preserve backward compatibility and make ownership explicit. Existing applications may return AutoCloseable objects whose lifecycle is managed elsewhere; automatically closing them would silently change current behavior and could cause premature or duplicate cleanup. DuckDBTableFunctionState provides an explicit opt-in contract while still following the familiar AutoCloseable API.
But if you think AutoClosable is enough, we can use it instead.

staticlibs:
I agree with the reasoning. Just I thought it may be enough to document, that `AutoClosable` will be closed by the native callback, for example for 2.0, can be revisited before the release.

PR #813, opened 2026-08-14
Add batched VARCHAR writes for table functions

lfkpoa:
### Motivation

Cell-by-cell VARCHAR writes repeatedly perform UTF-8 encoding, temporary allocations, JNI transitions, locking, and validation. Local smoke tests showed that batched writes improve throughput by approximately 2.4x-4.0x, depending on the data.

### Implementation

- Implement setStrings(long, String[]) for batched writes from Java strings.
- Reuse a contiguous UTF-8 payload buffer and metadata within each vector.
- Perform one JNI transition and one lock acquisition per column/chunk.
- Validate the complete payload once, then assign values in a native loop.
- Add setStringUtf8Batch(long, byte[][]) for producers that already have UTF-8 encoded values.
- Preserve null, empty-string, Unicode, long-string, and partial-chunk semantics.
- Keep the existing cell-by-cell API unchanged for compatibility.

### Files Changed
- src/main/java/org/duckdb/DuckDBWritableVector.java: implements setStrings and setStringUtf8Batch.
- src/main/java/org/duckdb/DuckDBBindings.java: declares the batch JNI binding.
- src/jni/bindings_vector.cpp: validates the combined UTF-8 payload and performs native batch assignment.
- duckdb_java.def, duckdb_java.exp, duckdb_java.map: export the new JNI symbol.
- src/test/java/org/duckdb/TestBindings.java: tests batch vector writes, validation, nulls, Unicode, and long strings.
- src/test/java/org/duckdb/TestTableFunctions.java: adds table-function equivalence and partial-chunk smoke tests.
- src/test/java/org/duckdb/TestStringWritePerformance.java: adds per-cell, String[] batch, and UTF-8 byte batch benchmarks.
- Makefile: adds the optional performance stress target.
- README.md: documents how to run the benchmark.

### Validation
- Full JDBC test suite passes.
- make format-check passes.
- Performance benchmark passes with assertions enabled.


src/jni/bindings_vector.cpp
```diff
@@ -207,6 +207,88 @@ JNIEXPORT void JNICALL Java_org_duckdb_DuckDBBindings_duckdb_1vector_1assign_1st
...
+ * Signature: (Ljava/nio/ByteBuffer;J[B[JII)V
+ *
+ * Batch variant of duckdb_vector_assign_string_element_len: assigns count
+ * concatenated UTF-8 strings starting at row index, in a single JNI call.
+ * lengths[i] >= 0 marks the byte length of element i, a negative length marks
+ * a NULL element. The complete payload is validated once before the loop uses
+ * the unsafe assignment primitive for each valid element.
+ */
+JNIEXPORT void JNICALL Java_org_duckdb_DuckDBBindings_duckdb_1vector_1assign_1string_1elements(
+    JNIEnv *env, jclass, jobject vector, jlong index, jbyteArray data, jlongArray lengths, jint count,
```
staticlibs: I wonder whether we can remove the `count` argument here and use the size of the `lengths` array instead?

src/jni/bindings_vector.cpp
```diff
@@ -207,6 +207,88 @@ JNIEXPORT void JNICALL Java_org_duckdb_DuckDBBindings_duckdb_1vector_1assign_1st
...
+	duckdb_vector vec = vector_buf_to_vector(env, vector);
+	if (env->ExceptionCheck()) {
+		return;
+	}
+	auto payload = make_jbyteArray_ptr(env, data);
+	if (env->ExceptionCheck()) {
+		return;
+	}
+	if (count < 0 || data_length < 0 || data_length > env->GetArrayLength(data)) {
+		env->ThrowNew(J_SQLException, "Invalid string batch length");
```
staticlibs, line 233: M, the exceptions handling in UDFs (and chunked result set) is not entirely consistent, but in general we try to not throw `SQLException` outside of the JDBC API functions (where it is mandated by the spec). Can we throw the `FunctionException` here (and below in this call) instead?

src/main/java/org/duckdb/DuckDBWritableVector.java
```diff
@@ -27,6 +32,8 @@ public final class DuckDBWritableVector {
     private final DuckDBVectorTypeInfo typeInfo;
     private final ByteBuffer data;
     private final ByteBuffer validity;
+    private byte[] batchPayload;
+    private CharsetEncoder utf8Encoder;
```
staticlibs: I wonder how heavy is this mutable state, can we just re-create batchPayload and utf8Encoder on every call, of will it be too expensive? In the latter case, lets add a comment here, that these too can only be accessed behind the lock.

staticlibs (commented):
Thanks for the PR! It looks good to me, added a few minor comments/questions.

src/jni/bindings_vector.cpp
lfkpoa: Accepted and implemented. The public Java batch setters now convert errors from the native call into DuckDBFunctions.FunctionException, keeping SQLException handling at the JDBC boundary. A regression test was added for invalid UTF-8 input.

src/main/java/org/duckdb/DuckDBWritableVector.java
lfkpoa: Accepted and implemented. Since there is one batch call per apply and the vector is recreated for each chunk, retained batchPayload/utf8Encoder state was not reused across apply calls. The state is now local to each batch invocation. A local smoke benchmark showed about 1.1x for ASCII values and about 1.0x for Unicode values, so retaining the state provided no material benefit.

src/jni/bindings_vector.cpp
lfkpoa: Accepted and implemented. The redundant count argument was removed from the JNI signature; the native binding now derives the number of elements from lengths.length, and the Java declaration/call site were updated accordingly.

staticlibs:
Thanks for the update!

PR #829, opened 2026-08-21
Implement RETURN_GENERATED_KEYS support via DML RETURNING rewrite

rmannibucau:
Background: using JDBC API with RETURN_GENERATED_KEYS in the original application totally broke with duckdb, this provides a fallback which should work most of the time

side note: no issue to close my PR if you want to implement it totally differently, I'm not ultra proud of the string parsing


src/main/java/org/duckdb/DuckDBPreparedStatement.java
```diff
@@ -1426,4 +1514,260 @@ private DirectQueryResult(ByteBuffer resultRef, DuckDBPendingQuery pendingQuery)
...
+        }
+
+        return autoDetectGeneratedColumns(table);
+    }
+
+    /**
+     * Extracts the (schema-qualified) target table name from a DML statement, quoted/unquoted.
+     */
+    private String extractTargetTable(String sql) {
+        String lowercase = sql;
```
staticlibs: I wonder if the lower case call supposed to be added here?

src/main/java/org/duckdb/DuckDBPreparedStatement.java
```diff
@@ -1426,4 +1514,260 @@ private DirectQueryResult(ByteBuffer resultRef, DuckDBPendingQuery pendingQuery)
...
         }
     }
+
+    /**
+     * Rewrites a DML statement ({@code INSERT}/{@code UPDATE}/{@code DELETE}) to append a
+     * {@code RETURNING <columns>} clause so that generated keys can be captured. If the statement
+     * cannot be reliably rewritten, or it already has a {@code RETURNING} clause, the original SQL is
+     * returned unchanged.
+     */
+    private String rewriteForReturning(String sql) throws SQLException {
```
staticlibs, line 1524: Nit: can we make all parsing helpers `static` where possible?

src/test/java/org/duckdb/TestPrepare.java
```diff
@@ -338,6 +338,298 @@ public static void test_execute_autogen_keys() throws Exception {
...
+                assertThrows(() -> s.execute("INSERT INTO t VALUES (1)", 42), SQLFeatureNotSupportedException.class);
+
+                assertThrows(
+                    () -> s.executeUpdate("INSERT INTO t VALUES (1)", 42), SQLFeatureNotSupportedException.class);
+            }
+        }
+    }
+
+    public static void test_generated_keys_invalid_flag_value_message() throws Exception {
+        try (Connection conn = DriverManager.getConnection(JDBC_URL); Statement s = conn.createStatement()) {
```
staticlibs: `s` triggers an "unused" warning, can we remove it?

staticlibs (approved):
Hi, thanks for the PR! It looks good to me, I added just one question and a couple of nits.

The parsing code may be not elegant, but I can see no other easy ways to implement it. We don't want to introduce real Java-level SQL parsers, as in future we are going to add a native parser extension to handle date time literals like `{d 'yyyy-mm-dd'}`. At that point going to look into changing current ad-hoc rewriting code to use a parser extension too.

rmannibucau:
tried to address the feedbacks

staticlibs:
Thanks for the update!

staticlibs:
Just FYI, snapshot with this change included (compatible with 1.5.5 extensions):

```xml
<dependency>
    <groupId>org.duckdb</groupId>
    <artifactId>duckdb_jdbc</artifactId>
    <version>1.5.5.2-dev-f4e6c80</version>
</dependency>
```
```xml
<repository>
    <id>duckdb</id>
    <url>https://duckdb-staging.duckdb.org/duckdb/duckdb-java/maven/</url>
</repository>
```

PR #836, opened 2026-08-24
[SQLException] fill error code and sql state for sql exceptions

rmannibucau:
(helped from - LLM  - deepseek v4 flash to classify and inject the categories/codes)

idea is to be able to identity the error from caller standpoint without parsing the message


staticlibs:
Thanks for the PR! I think this is a good addition, I will look into its details tomorrow.

src/main/java/org/duckdb/SQLState.java
```diff
@@ -0,0 +1,88 @@
...
+ * driver. Values follow the standard JDBC / SQLSTATE class taxonomy (first two characters select the
+ * SQLSTATE class, the last three the subclass) as listed in the IBM SQLSTATE value documentation and
+ * the JDBC 4.3 specification, extended where the driver needs a vendor-specific condition.
+ *
+ * <p>The state assigned to a given error is a function of the <em>origin</em> of the failure, not of
+ * the message text: closed-object errors map to {@code 08003}, data/type-conversion failures map into
+ * the {@code 22xxx} class, parameter problems map into {@code 0700x}, and everything else that has no
+ * precise SQL class defaults to {@link #HY000}.
+ */
+public enum SQLState {
```
staticlibs: Can we make the class and all its methods package private? It is not part of public API, may want to rename it in future.

src/main/java/org/duckdb/ErrorCode.java
```diff
@@ -0,0 +1,291 @@
...
+ * grouped by origin, and unique. The {@code getOrigin()} identifies the throwing class.
+ *
+ * <p>Native DuckDB errors (returned as free text by the C API) are wrapped by
+ * {@link JdbcUtils#createSQLExceptionFromNativeError(String, Throwable)} and receive
+ * {@link #NATIVE_UNDECODED}; their SQLState is derived from the error text prefix, see
+ * {@link JdbcUtils#nativeState(String)}.
+ *
+ * <p>This enum is the documented source of truth for error codes; see also {@code SQL_ERRORS.md}.
+ */
+public enum ErrorCode {
```
staticlibs: Can we make the class and all its methods package private? It is not part of public API, may want to rename it in future.

src/main/java/org/duckdb/ErrorCode.java
```diff
@@ -0,0 +1,291 @@
...
+        return origin;
+    }
+
+    /** Short human-readable description of the failure. */
+    public String getDescription() {
+        return description;
+    }
+
+    /** Builds a fully-populated {@link java.sql.SQLException} for this code. */
+    public java.sql.SQLException asException() {
```
staticlibs: We likely don't need fully qualified name here.

src/main/java/org/duckdb/ErrorCode.java
```diff
@@ -0,0 +1,291 @@
...
+            Integer block = blockByOrigin.get(ec.origin);
+            if (block == null) {
+                block = 1000 + blockByOrigin.size() * 100;
+                blockByOrigin.put(ec.origin, block);
+                seqByOrigin.put(ec.origin, 0);
+            }
+            int seq = seqByOrigin.get(ec.origin);
+            int code = block + seq;
+            seqByOrigin.put(ec.origin, seq + 1);
+            if (CODES.put(code, ec.name()) != null) {
```
staticlibs: Can we pre-fill this map and remove the static initializer?

src/main/java/org/duckdb/DuckDBAppender.java
```diff
@@ -132,7 +132,7 @@ public class DuckDBAppender implements AutoCloseable {
             if (null != appenderRef) {
                 duckdb_appender_destroy(appenderRef);
             }
-            throw new SQLException(createErrMsg(e.getMessage()), e);
+            throw JdbcUtils.createSQLException(createErrMsg(e.getMessage()), ErrorCode.APPENDER_CONVERSION, e);
```
staticlibs: Can we add a static import of `createSQLException` in all places where it is used?

src/main/java/org/duckdb/DuckDBAppender.java
```diff
@@ -1027,16 +1044,18 @@ private void moveToNextColumn() throws SQLException {
...
         if (ownerThreadId != 0 && ownerThreadId != currentThread().getId()) {
-            throw new SQLException(createErrMsg(
-                "detected the usage of the same Appender instance from multiple threads,"
-                + " owner thread ID: " + ownerThreadId + ", current thread ID: " + currentThread().getId() + ";"
-                + " 'append()' and 'flush()' operations cannot be called concurrently;"
-                + " when it is necessary to use the same Appender instance from multiple threads,"
-                + " call 'appender.unsafeBreakThreadConfinement()' method and use the 'Lock' instance"
-                + " obtained from there to synchronize the calls to the Appender."));
+            throw JdbcUtils.createSQLException(
+                createErrMsg("detected the usage of the same Appender instance from multiple threads,"
```
staticlibs, line 1051: I think two throwing helpers are unnecessary here, should be restructured to use a single helper.

staticlibs (approved):
The PR looks good to me! Added a few comments.

src/main/java/org/duckdb/io/IOUtils.java
```diff
@@ -5,6 +5,7 @@
 import java.io.InputStream;
 import java.io.Reader;
 import java.sql.SQLException;
+import org.duckdb.JdbcUtils;
```
staticlibs: Lets throw the errors as it was in the previous commit, without making JdbcUtils public.

src/main/java/org/duckdb/ErrorCode.java
```diff
@@ -0,0 +1,290 @@
...
+    BINDINGS_UNKNOWN_ID("DuckDBBindings", SQLState.DATA_EXCEPTION, "Invalid unknown ID"),
+    CHUNKED_RESULT_NATIVE("DuckDBChunkedResult", SQLState.HY000, "Query failed"),
+
+    /** Fallback for a native DuckDB error whose text prefix could not be classified. */
+    NATIVE_UNDECODED("duckdb.native", SQLState.HY000, "Undecoded native DuckDB error");
+
+    /** Code-to-name lookup pre-filled in {@link #allocateCodes()}. */
+    private static final Map<Integer, String> CODES = allocateCodes();
+
+    private static Map<Integer, String> allocateCodes() {
```
staticlibs: Perhaps I am missing something, but why do we need to generate numbers on class-load time? Why cannot these numbers be pre-computed?

src/main/java/org/duckdb/JdbcUtils.java
```diff
@@ -8,15 +8,16 @@
 import java.sql.SQLException;
 import java.util.Properties;
 
-final class JdbcUtils {
+public final class JdbcUtils {
```
staticlibs: Lets roll this change back, we don't want to add more public classes without reason.

src/main/java/org/duckdb/DuckDBAppender.java
```diff
@@ -132,7 +134,7 @@ public class DuckDBAppender implements AutoCloseable {
             if (null != appenderRef) {
                 duckdb_appender_destroy(appenderRef);
             }
-            throw new SQLException(createErrMsg(e.getMessage()), e);
+            throw createSQLException(createErrMsg(e.getMessage()), ErrorCode.APPENDER_CONVERSION, e);
```
staticlibs, line 136: Can we change this to use a single throw helper instead of 2 nested helpers?

staticlibs (changes requested):
Thanks for the update! Previously public classes are covered now. Its just static init in ErrorCode and nested helpers in Appender are holding this.

rmannibucau:
I hit some issues that i cant make all classes not public and use it everywhere so not sure which one to priviledge, i flattened some packages but not sure it is desired too

staticlibs:
> I hit some issues that i cant make all classes not public and use it everywhere so not sure which one to priviledge, i flattened some packages but not sure it is desired too

I suggest to special-case the classes in `io` and `user` subpackages (to not use throwing utils for them). And have package-private thrower for all classes in the root package.

rmannibucau:
are we good? anything I should update?

staticlibs:
Hi, thanks for the update! I wonder if you can address the remaining comments and fix the formatting (to allow CI to pass) by running:

```bash
python3 -m pip install --user clang_format==11.0.1
make format
```

Also, on the most recent change - can we add an overload to `createSQLException()` to not require callers to pass `null` when there is no `cause` exception?

rmannibucau:
hope it is better now

staticlibs:
@rmannibucau 

Thanks for the update! Would you mind adding the comments to each review point above describing the changes added for every point?

src/main/java/org/duckdb/SQLState.java
rmannibucau: Changed `public enum SQLState` to package-private `enum SQLState`. It is internal
   machinery, not part of the public API. All its methods were already package-private
   or private.

src/main/java/org/duckdb/ErrorCode.java
rmannibucau: Changed `public enum ErrorCode` to package-private `enum ErrorCode`, and made the
   previously-public accessors (`getCode()`, `getSQLState()`, `getOrigin()`,
   `getDescription()`, `asException()`, `count()`, `isUnique()`) package-private. The
   enum is only referenced from within the root `org.duckdb` package, so it can be
   renamed freely in the future without affecting users.

src/main/java/org/duckdb/ErrorCode.java
rmannibucau: Removed the static initializer and the `allocateCodes()` method that generated the
   code-to-name lookup on class-load. The numeric codes are now baked in as literal values
   on each enum constant (deterministic/stable, never computed at runtime). Lookup maps that
   are still needed are built lazily or from the literal values rather than in a static block.

src/main/java/org/duckdb/ErrorCode.java
rmannibucau: `asException()` no longer uses the `java.sql.SQLException` fully-qualified name; a plain
   `import java.sql.SQLException;` is used at the top of the file and `SQLException` is
   referenced unqualified.

src/main/java/org/duckdb/io/IOUtils.java
rmannibucau: Reverted IOUtils to throw plain `SQLException` / `new SQLException(e)` directly, exactly
   as before this PR. IOUtils no longer imports or calls into `JdbcUtils`.

src/main/java/org/duckdb/JdbcUtils.java
rmannibucau: Rolled back: `JdbcUtils` is again a package-private `final class` with package-private
   methods. `createSQLException` remains package-private and is used via static import from
   the classes in the root `org.duckdb` package.

src/main/java/org/duckdb/ErrorCode.java
rmannibucau: ack, no need

src/main/java/org/duckdb/DuckDBAppender.java
rmannibucau: Collapsed the two nested throwing helpers into one. All appender error sites now go through
   the single package-private `JdbcUtils.createSQLException(String, ErrorCode[, Throwable])`
   (imported statically), optionally wrapping the message with `createErrMsg(...)` where the
   existing duplicate-message behaviour must be preserved. No local throw helper remains in
   DuckDBAppender.

rmannibucau:
@staticlibs try to answer all cases but some were outdated so not sure how well it will render on your side

staticlibs:
@rmannibucau 

Thanks for the update! I think the only point left is about the Appender, it still uses 2 helpers to throw an error, like this:

```java
createSQLException(createErrMsg(...)...);
```

Can this be changed to use a single helper in Appender, not 2 helpers?

rmannibucau:
Another comment was tontry to avoir to pass null when there is not cause so added back the second flavors, do you have something different in mind or did I misread something?

staticlibs:
@rmannibucau 

I meant that it should be possible (at least I cannot see now why not) to combine `createSQLException` and `createErrMsg` helper in the Appender into a single helper. The PR is undoubtedly useful, and this untidy part in Appender is a really a minor part of it. Just I would like to not merge it in that unfinished state. I can complete it myself (keeping all the attribution on existing commits), just I don't have cycles to do that right now. So please feel free to leave the PR as is - I will return to it (to complete and merge it) in a few days.

rmannibucau:
did a pass I find not that bad as compromise, happy if you can follow up on this particular topic

src/main/java/org/duckdb/DuckDBAppender.java
rmannibucau: created helpers inlining both methods

PR #852, opened 2026-09-02
Register DuckDBTimestampTZ.getMicrosEpoch for JNI

geertjanw:
create_refs() looks up getMicrosEpoch via GetMethodID on the DuckDBTimestampTZ class during JNI_OnLoad, but the method is declared in its superclass DuckDBTimestamp. 

On HotSpot GetMethodID resolves inherited methods, so this works, but in a GraalVM native image the method must be registered on the queried class, so the lookup returned null, aborting JNI_OnLoad and preventing the driver from loading.

Register getMicrosEpoch on DuckDBTimestampTZ in the reachability metadata, and make the Graal native-image test's sample print the real exception, since the default uncaught exception handler can otherwise mask the cause in a native image.


staticlibs (approved):
Thanks for the PR! Looks good to me.

PR #854, opened 2026-09-02
Preserve native-image reachability metadata across vendoring

geertjanw:
The reachability-metadata.json JAR entry was added directly to the generated CMakeLists.txt but not to the CMakeLists.txt.in template. 

When the Vendor workflow regenerates CMakeLists.txt via vendor.py, the entry is dropped, so the driver JAR no longer bundles the GraalVM native-image metadata and the native-image build fails on the vendoring branch. 

This pr mirrors the entry into the template so it survives re-vendoring.


staticlibs:
Thanks!

